PR #8, opened 2021-01-08
fix fragment notice and URL

### LINKED ISSUE #4: job_fragment.yml hardcodes URL
via: https://github.com/nickdesaulniers/github_actions_playground/issues/1

https://github.com/nickdesaulniers/github_actions_playground/blob/a5cb81a33be488e2e0021a0e39937dcbf76eeb07/job_fragment.yml#L21

This will need to be retrieved from the generator.yml (repo/ref).

The fragment can be modified via the https://github.com/nickdesaulniers/github_actions_playground/blob/main/generate_workflow.py similar to what is done for the workflow name: https://github.com/nickdesaulniers/github_actions_playground/blob/a5cb81a33be488e2e0021a0e39937dcbf76eeb07/generate_workflow.py#L72

### LINKED ISSUE #7: job_fragment.yml hardcodes --tux-config argument
is causing the failure for the -next workflow.

nickdesaulniers:
Fixes #4
Fixes #7


nathanchance (approved):
LGTM.

PR #10, opened 2021-01-08
fix generating tuxbuild configs with llvm-(non-nightly)

### LINKED ISSUE #6: Handle llvm-* in addition to clang-*
Currently, `tuxmake`/`tuxmake` set LLVM=1 when using the `llvm-*` toolchain option, rather than `clang-*`. `generate_tuxbuild.py` hardcodes `llvm-nightly`, which has this as a TODO above it :)

https://github.com/ClangBuiltLinux/continuous-integration2/blob/7adb64944d4e8c99d12dc088c76bc60b95b90fb6/generate_tuxbuild.py#L53-L54

nickdesaulniers:
Add llvm-11 targets.

Fixes #6


nathanchance (approved):
This needs to be rebased on #8 because of the `.config` to `.yml` rename. Otherwise, LGTM.

PR #16, opened 2021-01-09
[RFC] Create a generated folder and add a small shell script for generating the command files

nathanchance:
The `generated` part of this PR is the RFC part. I think `generate.sh` is worthwhile regardless but it will need to be slightly modified if the `generated` folder is not popular :)


nickdesaulniers (approved):
bikeshed: the workflows are "generated" too, but must be in a separate directory.  Maybe "tuxbuild/", "tuxbuild_configs/", or "tux_configs/" would be more precise?

nathanchance:
Good point, although I suppose that kind of defeats the purpose of being obvious that "hey, these were made by another tool, don't touch them" but `tuxbuild/` makes sense. I will make that change shortly.

nathanchance:
Alright, move `generated/` to `tuxbuild/` and slightly modified the generated warning.

nickdesaulniers (approved):

PR #21, opened 2021-01-13
Use deterministic MD5 hash

nickdesaulniers:
previously could produce invalid names,
https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/481731328
errored with:
Names must start with a letter or '_' and contain only alphanumeric
characters, '-', or '_'.
It contained a `.` and `+` in -2.44522654963152E+18.

Also was not deterministic, so regenerating regenerated new hashes.

Prefix with `_`.

### DIFF
--- a/.github/workflows/next.yml
+++ b/.github/workflows/next.yml
@@ -18,7 +18,7 @@ jobs:
       with:
         path: builds.json
         name: output_artifact
-  -2445226549631522483:
+  _2f31caea64189172224e4f00071d26f1:
     runs-on: ubuntu-20.04
     needs: kick_tuxbuild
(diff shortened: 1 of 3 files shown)


.github/workflows/next.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       with:
         path: builds.json
         name: output_artifact
-  -2445226549631522483:
```
nickdesaulniers, line 21: oh, maybe these need to be string keys?

nickdesaulniers: A QA engineer walk into a bar, orders `-2445226549631522483` 🍻 🍻

msfjarvis (approved):
LGTM

PR #23, opened 2021-01-13
Set git_repo and git_ref in build-set

nathanchance:
Link: https://gitlab.com/Linaro/tuxbuild/-/blob/master/README.md#passing-git_repo-in-build-set


nathanchance:
@nickdesaulniers not sure if you were already working on it or not but is this what you had in mind? I think there is a lot of room for clean up here but I wanted to have a general idea.

generate_tuxbuild.py
```diff
@@ -20,31 +20,30 @@ def get_config():
...
     for build in config["builds"]:
-        if build["repo"] == repo and build["ref"] == ref:
+        if build["git_repo"] == repo and build["git_ref"] == ref:
             arch = build["ARCH"] if "ARCH" in build else "x86_64"
             toolchain = "clang-"
             toolchain += "nightly" if build["llvm_version"] == max_version else str(build["llvm_version"])
             kconfig = build["config"]
-            print("      - {{target_arch: {0}, toolchain: {1}, kconfig: {2}, make_variables: {{LLVM: \"{3}\"}}}}".format(\
-                    arch, toolchain, kconfig, int(build["llvm"])))
+            print("      - {{git_repo: \"{0}\", git_ref: \"{1}\", target_arch: {2}, toolchain: {3}, kconfig: {4}, make_variables: {{LLVM: \"{5}\"}}}}".format(\
```
nickdesaulniers, line 45: Can we avoid the escaped double quotes here? I suspect it's only needed when serializing an `int` as a `str`.

nickdesaulniers: Yeah, you can see it's adding `""` to the generated yml. We should be able to leave those off.

nathanchance: I am not sure, the documentation for tuxbuild shows them quoted. I did not try though.

nickdesaulniers (approved):

nickdesaulniers:
Is this still WIP?

nathanchance:
If you are happy with its current form, we can merge it and we can iterate on it further. Nothing is actually broken.

generate_tuxbuild.py
danrue: They no longer need to be quoted for tuxbuild - that was a bug and it was fixed..  Though in general this is easier to do if you create the data structure and then use print(yaml.dump()). Creating yaml (or json) using printed strings is harrowing!

PR #26, opened 2021-01-21
ci2: Enable some PowerPC builds

nathanchance:
We can enable all of the PowerPC builds for clang-12 because the issue
exposed by tuxmake has been fixed:

https://github.com/llvm/llvm-project/commit/0a23fbd28c7509f2f980946091e6055bf27164d2

We can enable ppc64le for clang-11 because it has the right target.

We cannot boot any of them because we do not have the right images.

### DIFF
--- a/.github/workflows/mainline.yml
+++ b/.github/workflows/mainline.yml
@@ -98,6 +98,66 @@ jobs:
         name: output_artifact
     - name: Boot Test
       run: ./check_logs.py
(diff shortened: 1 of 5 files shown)


nickdesaulniers (approved):
None "are the right images?"

nathanchance:
Actually, we could use `vmlinux.xz` for booting `pseries_defconfig` but we need to decompress it before booting it first. That will require some changes to `check_logs.py`, which I will do in a follow up commit, where I will enable booting.

PR #29, opened 2021-01-21
schedule cron jobs to run daily at midnight

nickdesaulniers:
We might be able to get fancier with these, based on how frequently
certain trees or toolchains get updated. For now, run something
continuously.

It seems that a whole workflow can only have one schedule...

### DIFF
--- a/.github/workflows/mainline.yml
+++ b/.github/workflows/mainline.yml
@@ -219,5 +219,7 @@ jobs:
     - name: Boot Test
       run: ./check_logs.py
 'on':
+  schedule:
+  - cron: 0 0 * * *
   workflow_dispatch: null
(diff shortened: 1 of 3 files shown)


nathanchance (approved):
LGTM.

nickdesaulniers:
It looks like this github action did not run overnight...

nickdesaulniers:
>     # * is a special character in YAML so you have to quote this string

via: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

nathanchance:
I think this is also because it is working off of UTC time and midnight UTC is 4pm UTC-8, which was after this was merged.

PR #32, opened 2021-01-21
Clang 11

nickdesaulniers:
check right LLVM version when boot testing
    
Requires LLVM_VERSION= env var to be set.


nathanchance (approved):
LGTM, make sure that this is regenerated on latest `main` before merging.

Issue #17, opened 2021-01-09
32-bit x86 should be using i386 for tuxbuild


nathanchance:
Using `x86_64` for the `target_arch` value with `i386_defconfig` as the `kconfig` argument results in an x86_64 kernel. Testing locally:

```
$ tuxmake --runtime podman --target-arch x86_64 --toolchain clang-nightly --kconfig i386_defconfig
...

$ rg 64BIT ~/.cache/tuxmake/builds/10/config
250:CONFIG_64BIT=y
807:CONFIG_PHYS_ADDR_T_64BIT=y
4406:CONFIG_ARCH_DMA_ADDR_T_64BIT=y
```

If `i386`  is used for `target_arch`, we at least run into https://github.com/ClangBuiltLinux/linux/issues/1210, which shows us that it works:

```
$ tuxmake --runtime podman --target-arch i386 --toolchain clang-nightly --kconfig defconfig
...
Unsupported relocation type: R_386_PLT32 (4)
...

$ rg 64BIT ~/.cache/tuxmake/builds/12/config
```

Testing with `clang-11` for the `toolchain`, we get a bootable kernel.

Ultimately, `generator.yml` should be updated to use `i386` for the `target_arch`.

PR #33, opened 2021-01-21
I386

nickdesaulniers:
i386 support
    
cc #17
    
Still need to add check_logs.py support once
https://github.com/ClangBuiltLinux/linux/issues/1210 is fixed.

### DIFF
--- a/.github/workflows/mainline.yml
+++ b/.github/workflows/mainline.yml
@@ -102,6 +102,27 @@ jobs:
         name: output_artifact
     - name: Boot Test
       run: ./check_logs.py
+  _c25e7cc6457b86e88a5b3379f84bab64:
+    runs-on: ubuntu-20.04
+    needs: kick_tuxsuite
+    name: ARCH=i386 LLVM=1 LLVM_IAS=1 BOOT=1 LLVM 12 defconfig
+    env:
+      ARCH: i386
+      LLVM: true
+      LLVM_VERSION: 12
+      LLVM_IAS: true
+      INSTALL_DEPS: 1
+      BOOT: true
+      CONFIG: defconfig
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - uses: actions/download-artifact@v2
+      with:
+        name: output_artifact
+    - name: Boot Test
+      run: ./check_logs.py
   _d23056385a8224ef80ba8d1f7b3c98a0:
     runs-on: ubuntu-20.04
     needs: kick_tuxsuite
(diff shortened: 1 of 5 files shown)


tuxbuild/mainline.tux.yml
```diff
@@ -6,13 +6,4 @@
...
-      - {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: arm, toolchain: clang-11, kconfig: aspeed_g5_defconfig, make_variables: {LLVM: "1"}}
-      - {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: arm, toolchain: clang-11, kconfig: multi_v7_defconfig, make_variables: {LLVM: "1"}}
-      - {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: arm64, toolchain: clang-11, kconfig: defconfig, make_variables: {LLVM: "1"}}
-      - {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: x86_64, toolchain: clang-11, kconfig: defconfig, make_variables: {LLVM: "1"}}
+      - {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: i386, toolchain: clang-nightly, kconfig: defconfig, make_variables: {LLVM: "1"}}
```
nathanchance, line 9: This diff doesn't look right?

nickdesaulniers: oh, yeah, I shouldn't have committed this...will undo, tomorrow

nickdesaulniers: (I was playing around with the file and invoking tuxsuite locally)

nickdesaulniers: done PTAL

nathanchance:
Does `utils.py` need to be updated to set `x86` for boot-utils?

https://github.com/ClangBuiltLinux/continuous-integration2/blob/2d6610767563026ce56d6763a45f1894b7de6383/utils.py#L58

Or should we just update boot-utils to take i386 instead of x86?

nickdesaulniers:
> Does utils.py need to be updated to set x86 for boot-utils?

Likely, see comment:

>> Still need to add check_logs.py support once
>> ClangBuiltLinux/linux#1210 is fixed.

(I don't expect this to close issue #17; it's just the first half. I'd prefer to fix up utils.py once the build actually produces a binary image remotely that I can test with)

nathanchance (approved):
Ah fair enough then, this is fine for now.

PR #36, opened 2021-01-21
Update boot-utils from upstream

msfjarvis:
Brings in https://github.com/ClangBuiltLinux/boot-utils/pull/31


nathanchance (approved):
Looks like we also get the s390 stuff for when we have a working QEMU binary.

PR #41, opened 2021-01-23
Greenify everything

nathanchance:


nathanchance:
This is mainly a proof of concept, I will rewrite `generator.yml` completely with these definitions if we are happy with them.

generator.yml
```diff
@@ -55,22 +55,29 @@ defconfigs:
   - &riscv    {config: defconfig,                 << : *riscv-triple}
   - &s390     {config: defconfig,                 << : *s390-triple}
   - &x86_64   {config: defconfig}
+tiers:
+  #                      LLVM=1       LLVM_IAS=1       Make variables to pass to TuxSuite
+  - &llvm_full          {llvm: true,  llvm_ias: true,  make_variables: {LLVM: 1, LLVM_IAS: 1}}
+  - &llvm               {llvm: true,  llvm_ias: false, make_variables: {LLVM: 1}}
+  - &clang              {llvm: false, llvm_ias: false, make_variables: {CC: clang}}
+  - &mips_llvm          {llvm: true,  llvm_ias: false, make_variables: {LD: mips-linux-gnu-ld, LLVM: 1}}
+  - &riscv_llvm_full    {llvm: true,  llvm_ias: true,  make_variables: {LD: riscv64-linux-gnu-ld, LLVM: 1, LLVM_IAS: 1}}
```
nickdesaulniers, line 64: so these would be:
```sh
$ ARCH=mips CROSS_COMPILE=mips-linux-gnu- make LLVM=1 LD=mips-linux-gnu-ld
$ ARCH=riscv CROSS_COMPILE=riscv64-linux-gnu- make LLVM=1 LLVM_IAS=1 LD=riscv64-linux-gnu-ld
```
?
(Does that work, I need to reread the Makefile's LLVM=1 support, I thought LD couldn't be specified on top of LLVM=1, but I haven't looked and don't remember).

nathanchance: Yes, it does. That is how we were doing things in the first CI repo. `LLVM=1` sets the default values but they can still be overridden individually. We could cut most things over to `LLVM=1` and just allowlist the GNU utils that we absolutely need.

nathanchance:
I have tried to logically chunk things up, the individual commits/messages should paint a nice picture and we get a successful TuxSuite run on both mainline and next :)

https://github.com/nathanchance/continuous-integration2/actions/runs/506217309
https://github.com/nathanchance/continuous-integration2/actions/runs/506217533

nathanchance:
With the fix to the `BOOT` value and adding i386 support, we now have two completely green runs:

mainline: https://github.com/nathanchance/continuous-integration2/actions/runs/506461732
next: https://github.com/nathanchance/continuous-integration2/actions/runs/506462059

nickdesaulniers (approved):
wow, thanks for all of this! 🚆 🔧

nathanchance:
Thanks for the review!

PR #45, opened 2021-01-24
check_logs.py: Clean up image and dtb location handling

nathanchance:
boot-qemu.sh supports booting an image with the full path to the image,
we do not need to fake a kernel directory.


nickdesaulniers (approved):
Very happy to see one less lookup table in utils.py.  🦘 🏴󠁧󠁢󠁷󠁬󠁳󠁿

Issue #43, opened 2021-01-24
stable jobs


nickdesaulniers:
Now that we have tuxsuite configs for mainline and next in place, let's extend these to stable and android common kernels.

https://github.com/ClangBuiltLinux/continuous-integration2/tree/main/tuxsuite
https://github.com/ClangBuiltLinux/continuous-integration2/tree/main/.github/workflows

PR #46, opened 2021-01-24
Add stable and Android trees

nathanchance:
Closes: #43

I will rebase this on #44 if we are happy with that so that we pick up the cron changes.


nathanchance:
Actually, this does not need to be rebased, the cron schedule is the same as in #44.

nickdesaulniers (approved):
I wish that commits:
- generate_tuxsuite.py: Add support for targets
- generator.yml: Add list of targets to build
were a distinct PR that went in first, but separating them out is needless busy work. LGTM!

🦉 🩹

PR #47, opened 2021-01-25
Add linting workflow

nathanchance:
I am not sure how this will feel but we do it on all of our other repos, I think that it is nice for consistency.


generate_workflow.py
```diff
@@ -43,23 +46,33 @@ def get_steps(build):
...
         sanitize_job_name(name): {
-            "runs-on": "ubuntu-20.04",
-            "needs": "kick_tuxsuite",
-            "name": name,
+            "runs-on":
+            "ubuntu-20.04",
+            "needs":
+            "kick_tuxsuite",
+            "name":
+            name,
```
nickdesaulniers, line 54: this change "hurts" but formatter knows best, I suppose...

nickdesaulniers (approved):
thanks! 🧼 🐌

generate_workflow.py
nathanchance: I can add `# yapf: disable` and `# yapf: enable` if we want to keep that the way it is for readability?

nickdesaulniers: sure, at the least we won't have to go look up the syntax for that every time we want to overrule the linter.

nickdesaulniers:
lgtm

PR #48, opened 2021-01-25
enable LLVM_IAS=1 for arm64/x86_64 back to 4.19

nickdesaulniers:
I've been working on this as part of Android S.

### DIFF
--- a/.github/workflows/4.19.yml
+++ b/.github/workflows/4.19.yml
@@ -37,10 +37,10 @@ jobs:
         name: output_artifact
     - name: Boot Test
       run: ./check_logs.py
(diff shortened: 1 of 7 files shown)


nathanchance (approved):
I wonder if we should enable the workflows to run when pull requests are made... Maybe not all of them, just the ones that are modified.

nickdesaulniers:
Do we have the ability to specify "only when changed" in the github actions yaml configs?

nathanchance:
Yes: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

See #50 in its current state.

PR #49, opened 2021-01-25
Improve TuxSuite command in yml header

nathanchance:


nickdesaulniers (approved):
yep, I noticed this last week. Should have filed a bug... 🐛 🐞

PR #50, opened 2021-01-25
WIP: Support running workflows when files are modified

nathanchance:
Currently, this will just run whenever the `yml` files are modified. Should we run all the workflows if we modify `check_logs.py` or `utils.py`?


nathanchance:
"Secrets are not passed to workflows that are triggered by a pull request from a fork."

:(

nickdesaulniers:
Good...otherwise someone could send us a PR that dumps our tokens

nathanchance:
I will open a new PR that changes this to a post submit style.

PR #52, opened 2021-01-25
Add post submit testing

nathanchance:
We can leverage GitHub's `push` trigger to do this cleanly.

See example (triggered by push): https://github.com/nathanchance/continuous-integration2/actions/runs/510480692


nickdesaulniers (approved):
LGTM

Though, let's say I send a PR where regen touches everything. I guess that's going to kick off a lot of builds?

nathanchance:
Yes, after it is merged.

Issue #34, opened 2021-01-21
Enable clang-nightly for certain targets


nathanchance:
Currently blocked on https://github.com/ClangBuiltLinux/linux/issues/1261.

nathanchance:
https://github.com/ClangBuiltLinux/linux/issues/1261 is fixed. We just have to wait for this patch to filter into the LLVM 12 packages from apt.llvm.org.

Issue #18, opened 2021-01-09
Enable RISCV builds


nathanchance:
This will be blocked on `tuxbuild` upstream: https://gitlab.com/Linaro/tuxbuild/-/issues/91

We need `LD=riscv64-linux-gnu-ld` because `ld.lld` does not support linker relaxation (https://github.com/ClangBuiltLinux/linux/issues/1020).

Additionally, we will need to `kconfig: [defconfig, "CONFIG_EFI=n"]` to workaround https://github.com/ClangBuiltLinux/linux/issues/1143.

nathanchance:
```
- {git_repo: "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git", git_ref: "master", target_arch: riscv, toolchain: clang-nightly, kconfig: [defconfig, CONFIG_EFI=n], make_variables: {LD: riscv64-linux-gnu-ld, LLVM: 1, LLVM_IAS: 1}}
```

works for the build set. However, I am not sure how that should be represented in `generator.yml`.

gctucker:
See also: https://groups.io/g/kernelci/message/994

As much as possible, it would be good to avoid duplicated efforts with KernelCI builds and other automated build system integrations.  Feel free to rely on the KernelCI Clang RISC-V builds as a reference if that helps, here's one:
https://kernelci.org/build/id/600a86edbd0ff54f40bb5d21/

It required this change in the Dockerfile: https://github.com/kernelci/kernelci-core/commit/1917b2ca0e3b4d44a27168303b9b6b2eb729bffd

We're gradually adding more architectures, let's hope we can keep sharing tips.

nathanchance:
https://github.com/nathanchance/continuous-integration2/commit/958a202c03771b98472a0af802477ce4866f5003

Issue #27, opened 2021-01-21
Enable MIPS and PowerPC booting


nathanchance:
Split off from #2.

Blocked on `tuxmake`: https://gitlab.com/Linaro/tuxmake/-/issues/109

nathanchance:
`tuxmake` has support now, which I have verified locally: https://gitlab.com/Linaro/tuxmake/-/merge_requests/100

We just need `tuxsuite` support now: https://gitlab.com/Linaro/tuxsuite/-/issues/95

PR #56, opened 2021-01-27
Fix and enable more architectures/configs

nathanchance:
All of these should be fairly self explanatory.

https://github.com/nathanchance/continuous-integration2/actions?query=branch%3Apresubmit%2Ffix-enable-more-stuff

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/18
Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/34


nickdesaulniers:
Would you mind separating out the CONFIG_EFI=n into its own pull request. I'm happy with everything else, but it's not clear precisely what's changed by that upon regen (at least I'd like to see it separately). In particular, I'm curious whether check_logs.py fetches the correct image to boot test in that case?

nathanchance:
We are not booting RISC-V right now: https://github.com/ClangBuiltLinux/continuous-integration2/issues/27

I can still separate that change out if you want me to but I verified that we select the right build based on the JSON dump: https://github.com/nathanchance/continuous-integration2/runs/1779320961?check_suite_focus=true

nickdesaulniers (approved):
looks like check_logs.py does work! 🆒

PR #61, opened 2021-01-28
Some small cleanups

nathanchance:
All of the commit messages should be pretty self explanatory.


nickdesaulniers (approved):
job names look nice. I kind of prefer having the comments about why builds are commented out directly above them, but no biggie. 😎

nathanchance:
I went ahead and dropped it, I was not entirely sold on it anyways :) thanks for the review!

PR #68, opened 2021-01-31
Add support for testing CONFIG_THUMB2_KERNEL

nathanchance:
All of the commits should be self explanatory and broken down properly.


utils.py
```diff
@@ -62,11 +62,12 @@ def get_requested_llvm_version():
...
     arch = os.environ["ARCH"]
-    config = os.environ["CONFIG"]
+    configs = sorted(os.environ["CONFIG"].split(","))
     llvm_version = get_requested_llvm_version()
     for build in _read_builds():
-        if build["target_arch"] == arch and build["kconfig"][0] == config \
-                and build["toolchain"] == llvm_version:
+        if build["target_arch"] == arch and \
+           build["toolchain"] == llvm_version and \
+           sorted(build["kconfig"]) == configs:
```
nickdesaulniers, line 70: how come we sort these configs?

nathanchance: Fair point, the lists are created in the same order so sorting shouldn't be necessary. I will test without that change later today.

gctucker:
BTW, KernelCI uses the `+` convention, and now you're inventing the `,` convention to achieve the same thing i.e. `multi_v7_defconfig,CONFIG_THUMB2=y` vs `multi_v7_defconfig+CONFIG_THUMB2=y`.  There's also a whole configuration mechanism to decide which kernel configs to build using YAML.  Maybe that doesn't fit your use-case, but I thought it was worth mentioning as it looked way too familiar to me.

nickdesaulniers:
That's fair feedback; @nathanchance it seems like if `+` was always used, we could split on that in utils.py (rather than `,`)? Or do we use `+` to separate other parts of the build string name?

nathanchance:
For what it is worth, I am only using the comma when internally passing around the list of configuration options between the workflows and `check_logs.py`/`utils.py`. We use "+" when showing the configuration options in the workflow name. I can change that to a "+" though because it would simplify the print config function to no longer need a delimiter parameter.

nathanchance:
Alright, that should look better, PTAL.

nickdesaulniers:
Are we keeping the sorting, or do you want to revisit that later?

nathanchance:
I need to explore if TuxSuite spits out the configuration options in the same way they were put in. If so, then dropping the sorted is fine but we are comparing our list against the one that is generated by TuxSuite and if TuxSuite reorders the configuration options, they won't match unless sorted. I don't think that is the case but I need to make sure.

nathanchance:
I think this should be ready to go, unless there are any other outstanding comments?

nickdesaulniers (approved):
yep, sorry, I had seen you'd dropped the sorting but got distracted before I could approve. Sorry for the delay! 🩹 📧

nathanchance:
Thanks for the review!

PR #71, opened 2021-02-03
enable LLVM_IAS=1 android x86 builds

nickdesaulniers:
two commits, one enables x86 since https://gitlab.com/Linaro/tuxmake/-/issues/117 is marked fixed. The other re-enables LLVM_IAS=1.

### DIFF
--- a/.github/workflows/android-4.19.yml
+++ b/.github/workflows/android-4.19.yml
@@ -50,4 +50,23 @@ jobs:
         name: output_artifact
     - name: Boot Test
       run: ./check_logs.py
+  _4fb2521d605de5401fc145b2b8a80bc8:
+    runs-on: ubuntu-20.04
+    needs: kick_tuxsuite_defconfigs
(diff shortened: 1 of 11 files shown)


nathanchance:
I just tested with `tuxmake` locally and I am not sure that this will pass: https://gitlab.com/Linaro/tuxmake/-/issues/117#note_500946481

I think 4.19 is going to still be broken regardless of that though: https://github.com/ClangBuiltLinux/linux/issues/1269

I will try to tackle that tomorrow.

nathanchance:
The `tuxmake` issue seems to be something on my end I guess. This needs to be rebased on latest main and have the workflow files regenerated. It is up to you if we disable x86_64 defconfig for 4.19, it will just be broken for the time being: https://builds.tuxbuild.com/1o1mS4y3G2efrRxXYaD48FVVPGP/

nathanchance (approved):
LGTM.

nickdesaulniers:
I'll follow up with disabling the known broken branch if we don't resolve the KI soon.

Issue #1, opened 2021-01-07
Enable MIPS builds


nathanchance:
This is blocked on `tuxmake` upstream: https://gitlab.com/Linaro/tuxmake/-/issues/110

nathanchance:
`tuxmake` has support now, which I have verified locally: https://gitlab.com/Linaro/tuxmake/-/merge_requests/100

We just need `tuxsuite` support now: https://gitlab.com/Linaro/tuxsuite/-/issues/95

Issue #72, opened 2021-02-04
RISC-V needs a newer version of QEMU


nathanchance:
From https://github.com/nathanchance/continuous-integration2/runs/1834110909, we run into a crash when shutting down the machine. As it turns out, I actually [fixed this in upstream QEMU](https://lore.kernel.org/qemu-devel/20200901055822.2721209-1-natechancellor@gmail.com/) but it seems like that patch never got pulled into Ubuntu's version. I think the easiest solution is to do what we did for s390x and package our own `qemu-system-riscv64`.

nathanchance:
Part one: https://github.com/ClangBuiltLinux/qemu-binaries/pull/2

Part two: https://github.com/ClangBuiltLinux/boot-utils/pull/33

Part three:

```diff
diff --git a/check_logs.py b/check_logs.py
index d21eaac..a9690e5 100755
--- a/check_logs.py
+++ b/check_logs.py
@@ -72,7 +72,7 @@ def run_boot():
     boot_qemu = [
         "./boot-utils/boot-qemu.sh", "-a", cbl_arch, "-k", kernel_image
     ]
-    if cbl_arch == "s390":
+    if cbl_arch == "riscv" or cbl_arch == "s390":
         boot_qemu += ["--use-cbl-qemu"]
     try:
         subprocess.run(boot_qemu, check=True)
```

PR #73, opened 2021-02-04
Enable booting MIPS and PowerPC kernels

nathanchance:
I will rebase this on #71 after it is merged but I wanted to get ahead of this for review.

TuxSuite has merged support for `kernel_image` so we can add that to `generator.yml` for architectures that need it.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/1
Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/27

### DIFF
--- a/generator.yml
+++ b/generator.yml
@@ -59,7 +59,7 @@ defconfigs:
   - &arm64_gki  {config: gki_defconfig,                                << : *arm64-triple,         << : *kernel_modules}
   - &arm64_cut  {config: cuttlefish_defconfig,                         << : *arm64-triple,         << : *kernel_modules}
   - &i386       {config: defconfig,                                    << : *i386-triple,          << : *kernel_modules}
-  - &mips       {config: malta_kvm_guest_defconfig,                    << : *mips-triple,          << : *kernel_modules}
+  - &mips       {config: [malta_kvm_guest_defconfig, CONFIG_CPU_BIG_ENDIAN=y],                    << : *mips-triple,          << : *kernel_modules}
   - &mipsel     {config: malta_kvm_guest_defconfig,                    << : *mipsel-triple,        << : *kernel_modules}
   - &ppc32      {config: ppc44x_defconfig,                             << : *powerpc-triple,       << : *kernel_modules}
   - &ppc64      {config: pseries_defconfig,                            << : *powerpc64-triple,     << : *kernel_modules}
--- a/generator.yml
+++ b/generator.yml
@@ -117,16 +117,12 @@ builds:
   - {<< : *arm32_v7_t, << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_tot}
   - {<< : *arm64,      << : *next,             << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
   - {<< : *i386,       << : *next,             << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
-  # mips and mipsel: Build disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/1)
-  # mips and mipsel: Boot disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/27)
-  # - {<< : *mips,       << : *next,             << : *mips_llvm,       boot: false, llvm_version: *llvm_tot}
-  # - {<< : *mipsel,     << : *next,             << : *llvm,            boot: false, llvm_version: *llvm_tot}
-  # ppc32, ppc64, and ppc64le: Boot disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/27)
-  - {<< : *ppc32,      << : *next,             << : *llvm,            boot: false, llvm_version: *llvm_tot}
-  - {<< : *ppc64,      << : *next,             << : *ppc64_llvm,      boot: false, llvm_version: *llvm_tot}
-  - {<< : *ppc64le,    << : *next,             << : *llvm,            boot: false, llvm_version: *llvm_tot}
-  # riscv: Boot disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/27)
-  - {<< : *riscv,      << : *next,             << : *riscv_llvm_full, boot: false, llvm_version: *llvm_tot}
+  - {<< : *mips,       << : *next,             << : *mips_llvm,       boot: true,  llvm_version: *llvm_tot}
+  - {<< : *mipsel,     << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_tot}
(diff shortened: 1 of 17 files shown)


nathanchance:
I do have RISC-V working already (https://github.com/ClangBuiltLinux/continuous-integration2/issues/72), which I can combine into this pull if it is desirable. Alternatively, we merge this and enable RISC-V in a separate pull.

generator.yml
```diff
@@ -100,16 +100,12 @@ builds:
...
-  - {<< : *ppc64,      << : *mainline,         << : *ppc64_llvm,      boot: false, llvm_version: *llvm_tot}
-  - {<< : *ppc64le,    << : *mainline,         << : *llvm,            boot: false, llvm_version: *llvm_tot}
-  # riscv: Boot disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/27)
-  - {<< : *riscv,      << : *mainline,         << : *riscv_llvm_full, boot: false, llvm_version: *llvm_tot}
+  - {<< : *mips,       << : *mainline,         << : *mips_llvm,       boot: true,  llvm_version: *llvm_tot}
+  - {<< : *mipsel,     << : *mainline,         << : *llvm,            boot: true,  llvm_version: *llvm_tot}
+  - {<< : *ppc32,      << : *mainline,         << : *llvm,            boot: true,  llvm_version: *llvm_tot}
+  - {<< : *ppc64,      << : *mainline,         << : *ppc64_llvm,      boot: true,  llvm_version: *llvm_tot}
+  - {<< : *ppc64le,    << : *mainline,         << : *llvm,            boot: true,  llvm_version: *llvm_tot}
+  - {<< : *riscv,      << : *mainline,         << : *riscv_llvm_full, boot: true,  llvm_version: *llvm_tot}
```
nickdesaulniers: will this depend on the riscv qemu changes?

generator.yml
```diff
@@ -59,13 +59,13 @@ defconfigs:
...
+  - &mips       {config: [malta_kvm_guest_defconfig, CONFIG_CPU_BIG_ENDIAN=y], kernel_image: vmlinux, << : *mips-triple,          << : *kernel_modules}
+  - &mipsel     {config: malta_kvm_guest_defconfig,                    kernel_image: vmlinux, << : *mipsel-triple,        << : *kernel_modules}
+  - &ppc32      {config: ppc44x_defconfig,                             kernel_image: uImage, << : *powerpc-triple,       << : *kernel_modules}
+  - &ppc64      {config: pseries_defconfig,                            kernel_image: vmlinux, << : *powerpc64-triple,     << : *kernel_modules}
+  - &ppc64le    {config: powernv_defconfig,                            kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
   #                                  https://github.com/ClangBuiltLinux/linux/issues/1143
-  - &riscv      {config: [defconfig, CONFIG_EFI=n],                    << : *riscv-triple,         << : *kernel_modules}
+  - &riscv      {config: [defconfig, CONFIG_EFI=n],                    kernel_image: Image, << : *riscv-triple,         << : *kernel_modules}
```
nickdesaulniers: Does specifying a `kernel_image` to tuxsuite prevent the kernel modules from being built?  (We don't want that; we want to build the kernel modules).  cc @danrue @terceiro

generator.yml
```diff
@@ -233,7 +237,8 @@ builds:
   # - {<< : *ppc32,      << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_latest}
   # - {<< : *ppc64,      << : *next,             << : *ppc64_llvm,      boot: true,  llvm_version: *llvm_latest}
   - {<< : *ppc64le,    << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_latest}
-  - {<< : *riscv,      << : *next,             << : *riscv_llvm_full, boot: true,  llvm_version: *llvm_latest}
+  # riscv: Boot disabled (https://github.com/ClangBuiltLinux/continuous-integration2/issues/72)
+  - {<< : *riscv,      << : *next,             << : *riscv_llvm_full, boot: false, llvm_version: *llvm_latest}
```
nickdesaulniers, line 241: maybe fold this commit into the one that enabled it? (Should be straightforward via interactive rebase)

install_deps.py
```diff
@@ -20,7 +20,7 @@ def install_deps():
         "powerpc": ["qemu-system-ppc"],
         "x86_64": ["qemu-system-x86"],
         "s390": [],
-        "riscv": ["qemu-system-riscv64"],
+        "riscv": ["opensbi", "qemu-system-riscv64"],
```
nickdesaulniers, line 23: this should wait until we enable riscv, separately.

nickdesaulniers (changes requested):
Let's keep the riscv changes distinct for now.

generator.yml
nickdesaulniers: (looks like yes)

nickdesaulniers:
I found it much easier to review multiple smaller commits in github's UI; thank for you for breaking up commits into smaller concerns.

generator.yml
nathanchance: The reason I did a separate commit was so that it was easy to revert when we wanted to re-enable RISC-V. Now that qemu-binaries and boot-utils have had their pulls merged, I could just update `check_logs.py` to have `--use-cbl-qemu` like I suggested in #72 and avoid disabling the builds altogether.

generator.yml
nathanchance: No. The modules are a separate target and we always build those.

generator.yml
nathanchance: Yes. See my other comment for why it might be advantageous to just enable booting RISC-V with this pull.

nathanchance:
> I found it much easier to review multiple smaller commits in github's UI; thank for you for breaking up commits into smaller concerns.

Good! I will continue to do this in the future.

generator.yml
nickdesaulniers: sure, so one more commit on top then?

nathanchance: Yes, basically :P

Please see https://github.com/nathanchance/continuous-integration2/commits/image-targets-with-riscv for what it would look like. If you think that is okay, I will push it under this branch so that the pull updates.

nickdesaulniers: LGTM

nathanchance:
Sigh that is not what I meant to do...

nickdesaulniers (approved):
nice, this gets us back to parity with all arches now, right? 🧇 🍟

nathanchance:
Yes, all of the architectures should boot if they can build :)

nathanchance:
Thanks for the review!

PR #74, opened 2021-02-05
ci2: Regenerate android-4.14 TuxSuite and workflow file

nathanchance:
This was missed by commit b907f6f ("regen").

PR #75, opened 2021-02-05
ci2: Add all*config builds

nathanchance:
NOTE: This is currently based on #74. I will rebase this after #73 and #74 are merged.

The general structure of this series:

1. Enable support for the `allconfigs` build-set in `generate_tuxsuite.py` and `generate_workflow.py`.
2. Add the anchors then format that section.
3. Enable the builds for mainline, next, and 5.10 then format those sections.
4. Regenerate the files.

If you would like anything broken down more, let me know.


generate_tuxsuite.py
```diff
@@ -68,8 +71,18 @@ def emit_tuxsuite_yml(config, tree):
             if "make_variables" in build:
                 current_build.update(
                     {"make_variables": build["make_variables"]})
-            tuxsuite_buildset["sets"][0]["builds"] += [current_build]
 
+            if "defconfig" in str(build["config"]):
```
nickdesaulniers, line 75: isn't `build["config"]` already a `str`? Why convert it again?

generator.yml
```diff
@@ -70,6 +70,18 @@ defconfigs:
...
+  - &arm32_mn_an {config: allnoconfig, << : *arm-triple, << : *kernel_modules}
+  - &arm32_mn_ay {config: allyesconfig, << : *arm-triple, << : *kernel_modules}
+  #                                      https://github.com/ClangBuiltLinux/linux/issues/1212
+  - &arm64_mn_am {config: [allmodconfig, CONFIG_UBSAN_UNSIGNED_OVERFLOW=n], << : *arm64-triple, << : *kernel_modules}
+  - &arm64_mn_an {config: allnoconfig, << : *arm64-triple, << : *kernel_modules}
+  - &arm64_mn_ay {config: [allyesconfig, CONFIG_UBSAN_UNSIGNED_OVERFLOW=n], << : *arm64-triple, << : *kernel_modules}
+  #                                       https://github.com/ClangBuiltLinux/linux/issues/256
+  - &x86_64_mn_am {config: [allmodconfig, CONFIG_UBSAN_UNSIGNED_OVERFLOW=n], << : *kernel_modules}
+  - &x86_64_mn_an {config: allnoconfig, << : *kernel_modules}
+  - &x86_64_mn_ay {config: [allyesconfig, CONFIG_UBSAN_UNSIGNED_OVERFLOW=n], << : *kernel_modules}
```
nickdesaulniers: These don't need to be in their own `allconfigs` section; can you put them in `defconfigs` (and retitle `defconfigs` to `configs` if it's more appropriate?

I'm not sure what `mn` is ("mainline").  Maybe `arm32_allyes`, `arm32_allno`, `arm32_allmod` is slightly more readable, if more verbose.

Noted on a later commit, but revisited this earlier commit:

I really really don't like the idea of conflating a config with a branch of a tree.  I'd rather have `all{no|yes|mod}config`s for each arch, not dependent on a tree.  I get this is trying to be precise due to https://github.com/ClangBuiltLinux/linux/issues/256, but my recommendation is to just disable `CONFIG_UBSAN_UNSIGNED_OVERFLOW` for `all{yes|mod}` for now, or come up with a way to factor out the fragment, such that the list of `defconfigs` is purely a config name (with optional fragment) and target triple.

generator.yml
```diff
@@ -82,6 +82,15 @@ allconfigs:
...
   - &x86_64_mn_ay {config: [allyesconfig, CONFIG_UBSAN_UNSIGNED_OVERFLOW=n], << : *kernel_modules}
+  - &arm32_nx_am {<< : *arm32_mn_am}
+  - &arm32_nx_an {<< : *arm32_mn_an}
+  - &arm32_nx_ay {<< : *arm32_mn_ay}
+  - &arm64_nx_am {<< : *arm64_mn_am}
+  - &arm64_nx_an {<< : *arm64_mn_an}
+  - &arm64_nx_ay {<< : *arm64_mn_ay}
+  - &x86_64_nx_am {<< : *x86_64_mn_am}
+  - &x86_64_nx_an {<< : *x86_64_mn_an}
+  - &x86_64_nx_ay {<< : *x86_64_mn_ay}
```
nickdesaulniers: YAGNI; add them when needed.

generator.yml
```diff
@@ -91,6 +91,16 @@ allconfigs:
...
+  #                                       https://git.kernel.org/linus/28187dc8ebd938d574edfc6d9e0f9c51c21ff3f4
+  - &arm32_510_am {config: [allmodconfig, CONFIG_CPU_BIG_ENDIAN=n], << : *arm-triple, << : *kernel_modules}
+  - &arm32_510_an {<< : *arm32_mn_an}
+  - &arm32_510_ay {config: [allmodconfig, CONFIG_CPU_BIG_ENDIAN=n], << : *arm-triple, << : *kernel_modules}
+  - &arm64_510_am {config: allmodconfig, << : *arm64-triple, << : *kernel_modules}
+  - &arm64_510_an {<< : *arm64_mn_an}
+  - &arm64_510_ay {config: allyesconfig, << : *arm64-triple, << : *kernel_modules}
+  - &x86_64_510_am {config: allmodconfig, << : *kernel_modules}
+  - &x86_64_510_an {<< : *x86_64_mn_an}
+  - &x86_64_510_ay {config: allyesconfig, << : *kernel_modules}
```
nickdesaulniers: I really think we should avoid distinct configs between different trees+branches.  This feels like a layering violation; suddenly our list of configs conflates a config with a certain branch of a certain tree.

generator.yml
```diff
@@ -91,6 +91,16 @@ allconfigs:
   - &x86_64_nx_am {<< : *x86_64_mn_am}
   - &x86_64_nx_an {<< : *x86_64_mn_an}
   - &x86_64_nx_ay {<< : *x86_64_mn_ay}
+  #                                       https://git.kernel.org/linus/28187dc8ebd938d574edfc6d9e0f9c51c21ff3f4
```
nickdesaulniers: I'd rather backport this to stable than have to have separate configs in our CI disabling BE.

nickdesaulniers (changes requested):
Let's see if there's simplifications to be had for the list of configs.

generate_tuxsuite.py
nathanchance: Not necessarily, this could be a list.

generator.yml
nathanchance: Sure, will delete.

generator.yml
nathanchance: Fair, I will request the backport then enable these targets when it is merged.

generator.yml
nathanchance: > These don't need to be in their own `allconfigs` section; can you put them in `defconfigs` (and retitle `defconfigs` to `configs` if it's more appropriate?

Sure, I can do that.

> I'm not sure what `mn` is ("mainline").  Maybe `arm32_allyes`, `arm32_allno`, `arm32_allmod` is slightly more readable, if more verbose.

Yes, mn is mainline, nx is next, and 510 is 5.10. I can make them more verbose.

> Noted on a later commit, but revisited this earlier commit:
> 
> 
> 
> I really really don't like the idea of conflating a config with a branch of a tree.  I'd rather have `all{no|yes|mod}config`s for each arch, not dependent on a tree.  I get this is trying to be precise due to https://github.com/ClangBuiltLinux/linux/issues/256, but my recommendation is to just disable `CONFIG_UBSAN_UNSIGNED_OVERFLOW` for `all{yes|mod}` for now, or come up with a way to factor out the fragment, such that the list of `defconfigs` is purely a config name (with optional fragment) and target triple.

That is fine by me. I just worry that one tree's target is going to break so then we disable the config for all of the trees and lose coverage. But at the same time, your suggestion is way simpler so we can go with that.

Realistically, we can solve that concern by having the ability to apply mbox files on top of the tree to work around issues, which is an outstanding tuxsuite request.

nickdesaulniers: Yes, the mbox file approach SGTM.

generator.yml
nathanchance: https://lore.kernel.org/r/20210205041351.GA2494386@localhost/

nathanchance:
Alright, I believe all of your comments should be addressed with this latest series. PTAL.

generate_tuxsuite.py
```diff
@@ -70,8 +73,18 @@ def emit_tuxsuite_yml(config, tree):
...
+                defconfigs += [current_build]
+            else:
+                allconfigs += [current_build]
```
nickdesaulniers, line 80: `.append` wouldn't create a new array

nickdesaulniers (approved):
Looks great, good work! 🥓 👟 heads up @danrue @terceiro for some allyesconfig and allmodconfig builds.

generate_tuxsuite.py
nathanchance: Are you talking about doing this change?

```diff
diff --git a/generate_tuxsuite.py b/generate_tuxsuite.py
index ecccea4..e4e8ad0 100755
--- a/generate_tuxsuite.py
+++ b/generate_tuxsuite.py
@@ -75,9 +75,9 @@ def emit_tuxsuite_yml(config, tree):
                     {"make_variables": build["make_variables"]})

             if "defconfig" in str(build["config"]):
-                defconfigs += [current_build]
+                defconfigs.append(current_build)
             else:
-                allconfigs += [current_build]
+                allconfigs.append(current_build)

     tuxsuite_buildset["sets"][0]["builds"] = defconfigs
     if allconfigs:
```

nickdesaulniers: yes

nathanchance: Done.

nathanchance:
Thanks for the review!

Issue #62, opened 2021-01-28
Handle LLVM 12 to LLVM 13 transition


nathanchance:
- [x] Wait for LLVM 12 binaries on [apt.llvm.org](https://apt.llvm.org/)
- [x] Report availability of LLVM 12 binaries on [apt.llvm.org](https://apt.llvm.org/) to [TuxMake issue tracker](https://gitlab.com/Linaro/tuxmake): https://gitlab.com/Linaro/tuxmake/-/issues/119
- [x] Wait for TuxMake to provide `clang-12` images: https://gitlab.com/Linaro/tuxmake/-/merge_requests/103
- [x] Wait for TuxSuite to allow `clang-12` to be specified: https://gitlab.com/Linaro/tuxsuite/-/issues/98
- [ ] Update `generator.yml` to provide `llvm_11` anchor, shift `llvm_latest` to 12, and shift `llvm_tot` to 13

Just a heads up @DanRue

danrue:
cc @terceiro

nickdesaulniers:
cc @sylvestre

dileks:
On Sat, Jan 30, 2021 at 7:37 AM Nick Desaulniers <notifications@github.com>
wrote:

> cc @sylvestre <https://github.com/sylvestre>
>

llvm-toolchain-12 and llvm-toolchain fired up...

# LC_ALL=C apt-cache policy clang-12

clang-12:
 Installed: 1:12.0.0~++20210131094231+07dc51637cc4-1~exp1~20210131084941.4
 Candidate: 1:12.0.0~++20210131094231+07dc51637cc4-1~exp1~20210131084941.4
 Version table:
*** 1:12.0.0~++20210131094231+07dc51637cc4-1~exp1~20210131084941.4 100
        99 https://apt.llvm.org/unstable llvm-toolchain-12/main amd64
Packages
       100 /var/lib/dpkg/status
    1:12~++20201105093023+cc2a2bb5ce5-1~exp1 99
        99 http://ftp.debian.org/debian experimental/main amd64 Packages
        99 https://deb.debian.org/debian experimental/main amd64 Packages

# LC_ALL=C apt-cache policy clang-13
clang-13:
 Installed: (none)
 Candidate: 1:13~++20210131111112+be997cead7fa-1~exp1~20210131101819.3757
 Version table:
    1:13~++20210131111112+be997cead7fa-1~exp1~20210131101819.3757 99
[...]

PR #79, opened 2021-02-09
llvm_tot is now 13

nathanchance:
Part one of https://github.com/ClangBuiltLinux/continuous-integration2/issues/62. There is no functional change, this is just matching reality.


nickdesaulniers (approved):
ok, let's follow up with @danrue @terceiro @sylvestre for llvm 12. 🧇 🍟

nathanchance:
Thanks for the review!

PR #82, opened 2021-02-10
Enable more some more configs on 5.10 and next

nathanchance:
The issues have been resolved so let's get the coverage while we can :)


generator.yml
```diff
@@ -305,9 +304,9 @@ builds:
   - {<< : *arm32_v7,      << : *stable-5_10,      << : *llvm,            boot: true,  llvm_version: *llvm_latest}
   - {<< : *arm32_v7_t,    << : *stable-5_10,      << : *llvm,            boot: true,  llvm_version: *llvm_latest}
   # arm32 all{mod,yes}config disabled (https://lore.kernel.org/r/20210205041351.GA2494386@localhost/)
```
nickdesaulniers, line 304: delete comment

nickdesaulniers (approved):
just a stale comment to remove. Thanks 👇 ☝️

generator.yml
nathanchance: Done. Thanks for catching it!

PR #83, opened 2021-02-10
Disable certain PowerPC builds with LLVM 11

nathanchance:
These should not have been enabled:

https://github.com/ClangBuiltLinux/continuous-integration2/runs/1874137744
https://github.com/ClangBuiltLinux/continuous-integration2/runs/1874137673


nickdesaulniers (approved):
💯

Issue #69, opened 2021-02-01
Add clang-cfi tree


nathanchance:
We should add testing for @samitolvanen 's LTO/CFI tree. We can build both the mainline and tip based branches. TuxSuite has the ability to specify the additional configurations needed on top of a configuration as a URL or in the tree under `kernel/configs/`: https://gitlab.com/Linaro/tuxsuite#kconfig-examples

Sami, if you wanted to, you could add an arm64 and x86_64 configuration fragment in that folder with the options you want us to test.

samitolvanen:
Sure. I'm currently running a CI locally, but more testing can't hurt. I'll take a look at adding the configs when I have a chance.

samitolvanen:
I think we should be fine with ThinLTO and CFI, and both arm64 and x86_64 can use the same config fragments. How does that look?

dileks:
On Wed, Feb 10, 2021 at 7:35 PM Sami Tolvanen <notifications@github.com>
wrote:

> I think we should be fine with ThinLTO and CFI, and both arm64 and x86_64
> can use the same config fragments. How does that look?
>

Last I looked at Clang-CFI it build and boot on bare metal on x86_64 - with
a slightly modified Debian-5.10 kernel-config for Linux v5.7-rcX.
So, GoGoGo.

nathanchance:
> I think we should be fine with ThinLTO and CFI, and both arm64 and x86_64 can use the same config fragments. How does that look?

Yes, this is exactly what I had in mind, thank you :) if you feel like any other configurations are worth testing (like if there is a config that has been problematic in combination with LTO/CFI in the past), feel free to add more configuration fragments and we can turn those into separate build targets.

I will take it from here getting your tree added to the matrix!

nathanchance:
@samitolvanen on your next push, would you mind adding that commit to the `tip/clang-cfi` branch as well? I'd like to add both branches since we have the capacity.

samitolvanen:
Sure, added also to tip/clang-cfi.

PR #84, opened 2021-02-10
Add Sami's LTO/CFI tree

nathanchance:
Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/69


nickdesaulniers (approved):
Can you add new badges to the readme for these trees?  Also, it seems like this is strictly only ThinLTO? That's fine, but it might be nice to make that more obvious than just the name of the config file?

samitolvanen (approved):

nathanchance:
> Can you add new badges to the readme for these trees?

Sure.

> Also, it seems like this is strictly only ThinLTO? That's fine, but it might be nice to make that more obvious than just the name of the config file?

Did you have any particular idea in mind? The logic for naming the workflows is to just pull from the options we fed to TuxSuite. I could move the configuration options into `generator.yml` rather than having the `kernel/configs/` files in Sami's tree but I figured that was a bit of a cleaner solution because then if the configuration options get renamed or there are other dependency configuration options that get pulled in, we can just update those config files rather than `generator.yml`. Crazy solution I suppose would be pulling down the `kernel/configs/` files and dumping them into the workflow name but that seems like overkill :)

generator.yml
```diff
@@ -245,6 +245,21 @@ builds:
   - {<< : *x86_64_cut,    << : *android-4_14,     << : *lld,             boot: true,  llvm_version: *llvm_tot}
   - {<< : *arm64_cut,     << : *android-4_9,      << : *clang,           boot: true,  llvm_version: *llvm_tot}
   - {<< : *x86_64_cut,    << : *android-4_9,      << : *clang,           boot: true,  llvm_version: *llvm_tot}
+  #############
+  #  LTO/CFI  #
+  #############
+  - {<< : *arm64,        << : *lto-cfi,          << : *llvm_full,        boot: true,  llvm_version: *llvm_tot}
+  - {<< : *arm64_lto,    << : *lto-cfi,          << : *llvm_full,        boot: true,  llvm_version: *llvm_tot}
```
nickdesaulniers, line 252: for example, when I see this anchor, I think "full LTO or ThinLTO" then have to go look at the anchor. not a big deal though.

nathanchance: Ah, I see. Fair point although I am not sure full LTO is worth testing through this matrix. Something to double back on if it is relevant in the future.

PR #85, opened 2021-02-10
Add support for arm64 big endian

nathanchance:
This cannot go in right away because TuxSuite's version of `ld.lld` is too old. I am merely pull requesting this for early review and I will merge it once that version of `ld.lld` is available.


nickdesaulniers (approved):
Do we need to upgrade our boot-utils git submodule too in order to boot the BE userspace image?

nathanchance:
Yes, thank you for the reminder!

nickdesaulniers:
👍

PR #86, opened 2021-02-10
Enable ld.lld for mips big endian

nathanchance:
This cannot go in right away because TuxSuite's version of `ld.lld` is too old. I am merely pull requesting this for early review and I will merge it once that version of `ld.lld` is available.


nickdesaulniers (approved):
LGTM 🥥 🐺

PR #92, opened 2021-02-24
Fix android-4.19

nathanchance:


nathanchance:
Looks like Sami pushed https://android-review.googlesource.com/c/kernel/common/+/1606628 so we can probably hold off on this for now.

nathanchance:
android-4.19 should pass now that https://android-review.googlesource.com/c/kernel/common/+/1606628 is merged.

PR #94, opened 2021-02-26
Enable LTO testing on mainline and -next

nathanchance:


samitolvanen (approved):

kees (approved):
LGTM. It might be good to also have allmodconfig+thinlto (defconfig is pretty small coverage). Any thoughts on including full LTO, or is that just asking for trouble? :)

nathanchance:
I can follow up on allmodconfig with ThinLTO, that seems worth testing and hopefully there should not be a huge increase in build time because of the parallelism that ThinLTO offers.

As for full LTO, that is probably going to be a question for @terceiro or @danrue, as it probably depends on if there are any memory restrictions with TuxSuite. I'll see what that consumes on my local machine so we have a benchmark.

danrue:
I don't anticipate any problems with memory, but I would be curious about build time and disk space requirements.

nathanchance:
ThinLTO does not seem to have a major impact on compile time with allmodconfig, although this test was on an AMD Epyc 7502P server, 64 threads so that might not translate to TuxSuite. I will run some tests locally to see.

Full LTO should probably be fine as well, maximum amount of RAM that I see used is 4GB by `ld.lld` during the final link.

nathanchance:
Whoops, forgot to upload the ThinLTO test results:

Screenshot 2021-02-27 110016

I will follow up on those in a separate PR later in the week.

PR #101, opened 2021-03-04
Thumb2 ias

nickdesaulniers:


nathanchance (approved):
LGTM, I tested mainline with `tuxmake` locally and the build is still green.

PR #103, opened 2021-03-05
utils.py: Fix get_cbl_name() for mipsel

nathanchance:
malta_defconfig is now the base, not malta_kvm_guest_defconfig.


nickdesaulniers (approved):
thanks for the fix.

I've been thinking about how I didn't want to pull in a python yaml library for the workers. I wonder if we should generate JSON from processing the generator.yml for tables like this, and write simple a simple json file the worker could parse. Then we could remove the hard coded llvm latest version number, with everything encoded in generator.yml.

PR #106, opened 2021-03-12
Move Android back to ThinLTO for now

nathanchance:
Addresses the recent failures with Android trees due to the move to full LTO.

### DIFF
--- a/.github/workflows/android-4.19.yml
+++ b/.github/workflows/android-4.19.yml
@@ -31,16 +31,16 @@ jobs:
       with:
         path: builds.json
         name: output_artifact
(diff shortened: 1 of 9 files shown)


vishalbhoj:
Let us wait on this patch. I have pushed a fix in tuxbuild as mentioned here:  https://github.com/ClangBuiltLinux/continuous-integration2/issues/105#issuecomment-797248486

nathanchance:
Thanks, I kicked the job again and all of the build succeeded (looks like GitHub Actions is having some issues so that is why the build stayed red).

https://github.com/ClangBuiltLinux/continuous-integration2/runs/2097106485

PR #107, opened 2021-03-12
Gcov

nickdesaulniers:


nathanchance:
This can be done later but I assume it is probably worth expanding this in terms of trees and architectures (mainline and arm64 maybe, given that the report came from Qualcomm?)

nathanchance (approved):
LGTM. Is this going to be green without your patch?

nickdesaulniers:
> This can be done later but I assume it is probably worth expanding this in terms of trees and architectures (mainline and arm64 maybe, given that the report came from Qualcomm?)

This bug wasn't arch specific.  There's diminishing returns for additional coverage IMO, though we do have the capacity. :)

> LGTM. Is this going to be green without your patch?

The build will be, but the boot wont be. So let's wait to merge until https://lore.kernel.org/lkml/20210312192139.2503087-1-ndesaulniers@google.com/T/#u is in -next?

nathanchance:
> > This can be done later but I assume it is probably worth expanding this in terms of trees and architectures (mainline and arm64 maybe, given that the report came from Qualcomm?)
> 
> This bug wasn't arch specific. There's diminishing returns for additional coverage IMO, though we do have the capacity. :)

Fair enough :)

> > LGTM. Is this going to be green without your patch?
> 
> The build will be, but the boot wont be. So let's wait to merge until https://lore.kernel.org/lkml/20210312192139.2503087-1-ndesaulniers@google.com/T/#u is in -next?

Yes, I think that is smart. I will test and review in a little bit.

nathanchance:
On its way to Linus, then we could turn this on for mainline and next?

https://lore.kernel.org/r/20210325043744.olufBBZcn%25akpm@linux-foundation.org/

nickdesaulniers:
> On its way to Linus, then we could turn this on for mainline and next?

This is a config that I suspect will have diminishing returns for the additional test coverage.  I may eat my words someday, but I'll add the CI coverage if so.

generator.yml
```diff
@@ -185,6 +186,7 @@ builds:
   - {<< : *x86_64_allmod_lto, << : *next,             << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
   - {<< : *x86_64_allno,      << : *next,             << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
   - {<< : *x86_64_allyes,     << : *next,             << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
+  - {<< : *x86_64_gcov,   << : *next,             << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
```
nickdesaulniers, line 189: d'oh, when rebasing, I fixed the indentation of the first hunk but not this.  Oh well.

PR #108, opened 2021-03-12
Add support for arm64 trees

nathanchance:


nickdesaulniers (approved):
great idea, thank you! Mind adding the badges to the gh-pages repo so they show up on the homepage, too?

generator.yml
```diff
@@ -33,6 +34,8 @@ trees:
   - &lto-cfi          {git_repo: *lto-cfi-url,  git_ref: clang-cfi,           name: lto-cfi,          << : *daily}
   - &lto-cfi-tip      {git_repo: *lto-cfi-url,  git_ref: tip/clang-cfi,       name: lto-cfi-tip,      << : *daily}
   - &tip              {git_repo: *tip-url,      git_ref: master,              name: tip,              << : *daily}
+  - &arm64-core       {git_repo: *arm64-url,    git_ref: for-next/core,       name: arm64,            << : *daily}
+  - &arm64-fixes      {git_repo: *arm64-url,    git_ref: for-next/fixes,      name: arm64-fixes,      << : *daily}
```
nickdesaulniers, line 38: @willdeacon I assume these two branches are appropriate for us to test?

nathanchance:
Yes, done: https://github.com/ClangBuiltLinux/ClangBuiltLinux.github.io/pull/21

generator.yml
willdeacon: Yes, although these two branches feed directly into linux-next, so if you're already testing that then I don't think we'd gain a tonne by testing these as well.

We do have a "for-kernelci" branch that gets updated less frequently, but does get picked up by kernelci and cki, so that might be a better option.

nathanchance: The idea of testing these trees in addition to linux-next is we can see when regressions spread across multiple trees or if they are localized to just one. We have the build capacity so getting as much coverage as possible is the goal.

I can change it to the `for-kernelci` branch though.

willdeacon: If testing in addition to linux-next is the goal, then stick with the two branches you have here. Given how rarely we push to `for-kernelci`, perhaps you could add that as a third target, but it's only worth testing it when `HEAD` has actually changed.

PR #110, opened 2021-03-23
Enable i386 builds on older trees

nathanchance:
The patch fixing the build has been backported to all supported trees.


nickdesaulniers (approved):
🧇 🍟

PR #112, opened 2021-03-24
boot-utils: Update

nathanchance:
Pulls in the latest rebuild of the images and the earlycon support.


nickdesaulniers (approved):
thanks! 💇‍♂️ 🏗️

PR #115, opened 2021-03-29
Add Android ARCH=arm LLVM_IAS=1 5.4+ coverage

nickdesaulniers:
add thumb2 coverage as well.

PR #116, opened 2021-03-30
Arm android cleanup

nickdesaulniers:
see also:
https://github.com/ClangBuiltLinux/linux/issues/1337
we could spin this down, but I'd rather have some coverage and work towards enabling LLVM_IAS=1/llvm_full for them.

Also, regardless of
https://github.com/ClangBuiltLinux/continuous-integration2/pull/115#issuecomment-810559773
I think we'll hit:
https://github.com/ClangBuiltLinux/linux/issues/1271

I have personal obligations today that I need to run to, I suggest merging these for tonight's CI run, then I'll clean up the rest tomorrow.

### DIFF
--- a/.github/workflows/android-5.10.yml
+++ b/.github/workflows/android-5.10.yml
@@ -88,10 +88,10 @@ jobs:
         name: output_artifact
     - name: Boot Test
       run: ./check_logs.py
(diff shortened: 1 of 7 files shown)


nathanchance (approved):
LGTM.

PR #120, opened 2021-04-20
add Android13 branch

nickdesaulniers:
moved the android12 branches around so that it's more obvious the difference between android12 branches and android13 branches.


nathanchance (approved):
Nice! LGTM.

PR #123, opened 2021-04-26
Make it green as of 4-26-2021

nathanchance:
This should make all of the CI green; the x86 failures on -next are not addressed by this because the problem has been fixed in the KVM tree, it should be fixed when linux-next is updated tonight.

The patches should make sense individually, let me know if there are any problems.


nickdesaulniers (approved):
thanks for untangling these! 🥥 ⛳

nathanchance:
Thanks for the review!

PR #126, opened 2021-04-30
Enable LLVM_IAS=1 for arm32 allmodconfig on android-mainline

nathanchance:
Android has started merging the 5.13 merges into android-mainline, meaning that we have the crypto fix that we need.


nathanchance:
Ping :)

nickdesaulniers (approved):
thanks for the ping, I must have dropped this from my notifications by accident; perhaps via checking email.  Pings work well for me. :) 🪓 💪

PR #127, opened 2021-05-03
Disable boot for ppc32 on mainline

nathanchance:
Same as commit 499feb7 ("generator.yml: Disable boot for ppc32 on
-next") but for mainline, now that the commit that broke booting has
been merged. This can be enabled again once we have a fix on the LLVM
side.


nickdesaulniers (approved):
thanks! 🥥 ⛳

PR #130, opened 2021-05-12
Add problem matcher

nathanchance:
The problem matcher allows us to get clear visuals of errors/warnings:

https://github.com/actions/toolkit/blob/e9c6ee99a59ec34586d4374b68efe5cabfcee9db/docs/problem-matchers.md

https://github.com/linuxppc/linux-ci/actions/runs/826137942

Do note, I have not actually tested this but this is how it is wired up elsewhere.

### DIFF
--- a/.github/problem-matchers/clang-errors-warnings.json
+++ b/.github/problem-matchers/clang-errors-warnings.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "clang-errors-warnings",
+            "pattern": [
+                {
+                    "regexp": "^(?:/builds/)?(.*):(\\d+):(\\d+):\\s+(?:fatal\\s+)?(warning|error):\\s+(.*)$",
+                    "file": 1,
+                    "line": 2,
+                    "column": 3,
+                    "severity": 4,
+                    "message": 5
+                }
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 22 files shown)


.github/problem-matchers/clang-errors-warnings.json
```diff
@@ -0,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "clang-errors-warnings",
+            "pattern": [
+                {
+                    "regexp": "^(?:/builds/)?(.*):(\\d+):(\\d+):\\s+(?:fatal\\s+)?(warning|error):\\s+(.*)$",
```
nickdesaulniers, line 7: Perhaps we want to match uppercase `WARNING` as well (ie. from modpost).

Is it intentional that some of these groups use matching, and others don't (`?:`)?

What is `/builds/`? Does that string appear in kernel build logs?

nathanchance: > Perhaps we want to match uppercase `WARNING` as well (ie. from modpost).

modpost warning are not the same pattern though so they won't match. I do eventually want to make problem matchers for all of these types of problems though.


> Is it intentional that some of these groups use matching, and others don't (`?:`)?

Yes, I believe so.

> What is `/builds/`? Does that string appear in kernel build logs?

Our source is mounted at `/builds/linux` within TuxSuite so that is the base of our warnings.

nickdesaulniers (approved):

PR #132, opened 2021-05-12
Build arm64 with CFI on mainline and next

nathanchance:
Setting to draft for now as -next is broken and I do not want to introduce more breakage: https://lore.kernel.org/r/CA+G9fYsRZswjeDZj1hiGDu5HZ+jYCVsOOL6MAzEb=LVfQ1_NpA@mail.gmail.com/

@samitolvanen: Once this is merged, you can remove the commit that adds the `.config` fragments unless you find them useful.


nickdesaulniers (approved):
thanks for the update! 🌴 🎋

PR #135, opened 2021-05-12
Wire up builds with AOSP LLVM

nathanchance:


utils.py
```diff
@@ -65,7 +62,10 @@ def _read_builds():
 
 def get_requested_llvm_version():
     ver = int(os.environ["LLVM_VERSION"])
-    return "clang-" + ("nightly" if ver == TOT_LLVM else str(ver))
+    ci_folder = pathlib.Path(__file__).resolve().parent
+    with open(ci_folder.joinpath("LLVM_TOT_VERSION")) as f:
+        llvm_tot_version = int(f.read())
+    return "clang-" + ("nightly" if ver == llvm_tot_version else str(ver))
```
stephenhines, line 68: Will this call AOSP Clang "clang-nightly" once we update again? We always pick up the version number from upstream, which moved to 13 recently.

nathanchance: No. Here is some backstory for this block.

The `ver` variable value comes from the `LLVM_VERSION` variable, which comes from our GitHub workflow files:

https://github.com/ClangBuiltLinux/continuous-integration2/blob/30da0066f38b5fd4dbb0938c1177b3dc641a2a71/.github/workflows/5.4.yml#L37/

Our generator defines `llvm_tot` as `13` but TuxSuite does not have a `clang-13` target, just `clang-nightly` to notate that those versions are coming from https://apt.llvm.org/ so we have to convert that `13` into `nightly`. When generating the workflow files for GitHub Actions, we use `LLVM_VERSION=13` instead of `LLVM_VERSION=nightly` because it is easier to understand but to find the build, we have to convert that `13` back into `nightly`.

For AOSP clang builds, `LLVM_VERSION` will just set to `android` because TuxSuite's name for AOSP clang is just `clang-android` since it only supports the latest version of AOSP clang.

stephenhines (approved):
Thanks for the extra info. The rest of the config looks great.

nickdesaulniers (approved):
great work, very happy to see a the much less brittle way of tracking ToT LLVM version! 🆒 🏖️

PR #137, opened 2021-05-13
utils.py: Fix crash with LLVM_VERSION=android

nathanchance:
```
  File "./check_logs.py", line 102, in <module>
    build = get_build()
  File
"/home/runner/work/continuous-integration2/continuous-integration2/utils.py",
line 74, in get_build
    llvm_version = get_requested_llvm_version()
  File
"/home/runner/work/continuous-integration2/continuous-integration2/utils.py",
line 64, in get_requested_llvm_version
    ver = int(os.environ["LLVM_VERSION"])
ValueError: invalid literal for int() with base 10: 'android'
```

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -61,11 +61,11 @@ def _read_builds():
 
 
 def get_requested_llvm_version():
-    ver = int(os.environ["LLVM_VERSION"])
+    ver = os.environ["LLVM_VERSION"]
     ci_folder = pathlib.Path(__file__).resolve().parent
     with open(ci_folder.joinpath("LLVM_TOT_VERSION")) as f:
-        llvm_tot_version = int(f.read())
-    return "clang-" + ("nightly" if ver == llvm_tot_version else str(ver))
+        llvm_tot_version = str(int(f.read())).strip()
+    return "clang-" + ("nightly" if ver == llvm_tot_version else ver)
 
 
 def get_build():


utils.py
```diff
@@ -61,11 +61,11 @@ def _read_builds():
 
 
 def get_requested_llvm_version():
-    ver = int(os.environ["LLVM_VERSION"])
+    ver = str(os.environ["LLVM_VERSION"])
```
nickdesaulniers, line 64: env vars are strings already, aren't they? unnecessary case?

utils.py
```diff
@@ -61,11 +61,11 @@ def _read_builds():
...
     ci_folder = pathlib.Path(__file__).resolve().parent
     with open(ci_folder.joinpath("LLVM_TOT_VERSION")) as f:
-        llvm_tot_version = int(f.read())
-    return "clang-" + ("nightly" if ver == llvm_tot_version else str(ver))
+        llvm_tot_version = str(f.read()).strip()
```
nickdesaulniers, line 67: perhaps still worth validating that the contents of this file can be converted to `int`?

utils.py
nathanchance: Yes, good call.

utils.py
nathanchance: ```llvm_tot_version = str(int(f.read())).strip()```?

Comparing `int` to `str` won't work.

nickdesaulniers: yes; that way if we put something that can't be parsed as an int, it will fail loudly (louder? maybe doesn't matter then)

nathanchance: Done.

nickdesaulniers (approved):

PR #140, opened 2021-05-20
Fixup the Hexagon jobs and optimize around BOOT=0

nathanchance:
`check_logs.py` fails currently due to `ARCH=hexagon` not being handled. At the same time, I noticed that we should not have even run into this because `BOOT=0` means we should not even bother installing dependencies, which is the second commit.


nickdesaulniers (approved):
thanks for the fix and great idea! 🏄 🙃

PR #142, opened 2021-05-24
generator.yml: Disable arm64 big endian build with AOSP LLVM

nathanchance:
AOSP LLVM does not have the patches that enable us to build arm64 with
LLVM=1 + CONFIG_CPU_BIG_ENDIAN=y, which were merged in LLVM 13:

https://github.com/llvm/llvm-project/commit/eea34aae2e74e9b6fbdd5b95f479bc7f397bf387
https://github.com/llvm/llvm-project/commit/7605a9a009b5fa3bdac07e3131c8d82f6d08feb7

On the kernel side, we restricted CONFIG_CPU_BIG_ENDIAN=y to LLVM 13:

https://git.kernel.org/linus/e9c6deee00e9197e75cd6aa0d265d3d45bd7cc28

As a result, we try to select CONFIG_CPU_BIG_ENDIAN but Kconfig makes us
silently fall back to CONFIG_CPU_LITTLE_ENDIAN=y, resulting in the
userspace boot failure seen here:

https://github.com/ClangBuiltLinux/continuous-integration2/runs/2655483207?check_suite_focus=true

Disable this build until AOSP LLVM does a new build based on LLVM 13
with the above fixes.

### DIFF
--- a/.github/workflows/next.yml
+++ b/.github/workflows/next.yml
@@ -1543,27 +1543,6 @@ jobs:
       run: echo "::add-matcher::.github/problem-matchers/clang-errors-warnings.json"
     - name: Boot Test
       run: ./check_logs.py
-  _66d5f7613f33d6225179b4f74b52c782:
-    runs-on: ubuntu-20.04
-    needs: kick_tuxsuite_defconfigs
-    name: ARCH=arm64 LLVM=1 LLVM_IAS=1 LLVM_VERSION=android defconfig+CONFIG_CPU_BIG_ENDIAN=y
-    env:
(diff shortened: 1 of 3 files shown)


nickdesaulniers (commented):
can you remove the `Closes` tag from the commit message? Let's keep that issue open for now?

Otherwise let's just delete this config

nathanchance:
Sure, I was going to open a new issue but it makes sense to just reuse that one.

nickdesaulniers (approved):
🇦🇽 🇦🇨

PR #145, opened 2021-06-02
Deal with most of the current fires

nathanchance:
This should resolve the vast majority of consistent failures over the past week or so.

The CI is pretty red for LLVM 13 because of an upstream issue, which [I reverted](https://github.com/llvm/llvm-project/commit/e6b086bef2c0597ed14b2aefbb3f6d74b94fd49e) but Debian's apt.llvm.org binaries built between the original patch and my revert :/ that should go away soon so it is not being dealt with here.

I'll ping the Hexagon patches today, hopefully that will be the first thing to get reverted.


nickdesaulniers (approved):
thanks for the cleanup! 🧹 🚽

PR #146, opened 2021-06-02
Enable pseries_defconfig (ppc64) again

nathanchance:
This was disabled because of [a `-Werror` issue](https://github.com/ClangBuiltLinux/linux/issues/1292). We disable -Werror for all
PowerPC builds because of a macro collision so we can re-enable this
build as well.

This matches PowerPC's own CI:

https://git.kernel.org/powerpc/c/62e2871dfaf347501cf3fa6a999acd48679088c2


nickdesaulniers (commented):
Nice find. Indeed especially if the PPC maintainer is doing this as well then I think we should be, too. 🚎 👍

nickdesaulniers (approved):

PR #149, opened 2021-06-17
Re-support RISCV defconfig with LLVM=1 LLVM_IAS=1

nickdesaulniers:
Thanks to @compnerd's fix:
https://reviews.llvm.org/rGbbea64250f65480d787e1c5ff45c4de3ec2dcda8
ClangBuiltLinux/linux#1023
ClangBuiltLinux/linux#1043

LLVM 13 (ToT) can now fully build RISCV defconfig again.

Split off the configs into riscv (defconfig) and riscv_no_efi. Older
versions of llvm should use riscv_no_efi.

Then keep the old riscv_llvm_full "tier," continue to use it for older
versions of llvm, but move llvm_tot back to use llvm_full.

### DIFF
--- a/.github/workflows/5.10.yml
+++ b/.github/workflows/5.10.yml
@@ -283,16 +283,16 @@ jobs:
       run: echo "::add-matcher::.github/problem-matchers/clang-errors-warnings.json"
     - name: Boot Test
       run: ./check_logs.py
-  _ea24606870e91e9589c8b50eea55772a:
+  _4908c92c4d725c03f7b7115b6b457e9e:
     runs-on: ubuntu-20.04
     needs: kick_tuxsuite_defconfigs
-    name: ARCH=riscv LLVM=1 LD=riscv64-linux-gnu-ld LLVM_IAS=1 LLVM_VERSION=13 defconfig+CONFIG_EFI=n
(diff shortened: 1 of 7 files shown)


compnerd (approved):

msfjarvis (approved):

nathanchance:
LGTM but it would have been nice to hold off on merging this until https://apt.llvm.org was updated with that fix to avoid breaking the build further.

nathanchance:
Also, I am not sure the 5.10 change is right since https://git.kernel.org/linus/ec3a5cb61146c91f0f7dcec8b7e7157a4879a9ee does not exist there?

nickdesaulniers:
> Also, I am not sure the 5.10 change is right since https://git.kernel.org/linus/ec3a5cb61146c91f0f7dcec8b7e7157a4879a9ee does not exist there?

Then https://git.kernel.org/linus/ec3a5cb61146c91f0f7dcec8b7e7157a4879a9ee should be sent to stable since it wasn't tagged. (cc @kraj )

nickdesaulniers:
5.10 should be fine because the "tier" is still using GNU BFD to link, right?

nathanchance:
> 5.10 should be fine because the "tier" is still using GNU BFD to link, right?

No, the tier moved from `riscv_llvm_full` to just `llvm_full`, which removed GNU ld.

PR #150, opened 2021-06-17
Disable s390 builds for clang 12 and earlier on -next

nathanchance:
This is no longer expected to work:

https://lore.kernel.org/r/YMtib5hKVyNknZt3@osiris/


nickdesaulniers (approved):
RIP 💀 🧟

PR #153, opened 2021-06-25
Update the start time of the -next jobs

nathanchance:
Towards the end of a release cycle, -next releases get later and later
since there are more merge conflicts and build errors to fight through,
meaning that with our current schedule, we often miss the version of
-next that is going to be released.

Push the time we start the builds by three hours, which should ensure
that -next is released by that time.

### DIFF
--- a/.github/workflows/next.yml
+++ b/.github/workflows/next.yml
@@ -13,7 +13,7 @@ name: next
     - tuxsuite/next.tux.yml
     - .github/workflows/next.yml
   schedule:
-  - cron: 0 8 * * 1,2,3,4,5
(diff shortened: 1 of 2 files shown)


nickdesaulniers (approved):
Can you help me verify how you determined 11:00 UTC? 🤷‍♂️ 🍤

nathanchance:
linux-next has been getting released at around 9pm Australian Eastern Standard Time (AEST), which is ten hours ahead of UTC, so that will be around 11am UTC. I should actually do noon just to make sure we catch it in case it goes even later than that.

nickdesaulniers (approved):

PR #155, opened 2021-06-29
check_logs.py: Do not use number of errors to fail the build

nathanchance:
It is possible for there to be errors in the build that are not fatal:

https://builds.tuxbuild.com/1uccc7bvODo3jO9zwLBJBhrudAw/build.log

https://lore.kernel.org/r/20210625122902.4058783-1-stefanb@linux.vnet.ibm.com/

The result field in status.json will tell us if the build fatally
errored based on make's error status so use that instead so that we do
not have false positive failures.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -83,8 +83,8 @@ def run_boot():
 
 
 def boot_test(build):
-    if build["errors_count"] > 0:
-        print_red("errors encountered during build, skipping boot")
+    if build["result"] == "fail":
+        print_red("fatal build errors encountered during build, skipping boot")
         sys.exit(1)
     if "BOOT" in os.environ and os.environ["BOOT"] == "0":
         print_yellow("boot test disabled via config, skipping boot")


check_logs.py
```diff
@@ -83,8 +83,8 @@ def run_boot():
 
 
 def boot_test(build):
-    if build["errors_count"] > 0:
-        print_red("errors encountered during build, skipping boot")
+    if build["result"] == "fail":
```
nickdesaulniers, line 86: I wonder why `"result"` differs from `"build_status"`? Would `"build_status"` be better to use?

nickdesaulniers (approved):

check_logs.py
nathanchance: @terceiro could you give us some insight on the difference between `"build_status"` and `"result"`? Presumably, `"build_status"` purely referred to the `make` pipeline whereas `"result"` refers to the entire `tuxsuite` pipeline? In other words, if there was some sort of infrastructure error, it is possible that `"result"` could be `"fail"` but `"build_status"` could be `"pass"`? If that is the case, it seems like `"result"` would still be the better one to use?

nathanchance:
I am going to merge this for now. If we should use `"build_status"` instead, I will do that in a follow up commit but I want the build for `-next` to be mostly green tomorrow.

check_logs.py
terceiro: I'm not sure, asked my colleagues to chime in

vishalbhoj: historically we had the below fields:
build_status: reflecting the status of the build(tuxmake) itself taking the values : pass/fail/error (error reflects infrastructure failure)
tuxbuild_status: reflecting the status of the infrastructure itself taking the values: provisioning/building/complete/error

As part of unified states of builds/test we introduced:
result: reflecting the status of the the build(tuxmake) itself taking the values: pass/fail/error (error reflects infrastructure error)
state: reflecting the status of the infrastructure itself taking the values: provisioning/running/finished
 
For  backward compatibility we still have the build_status and tuxbuild_status fields. You should use "result"  field to determine the failure and can later check for error_counts if needed.

nickdesaulniers: Thanks @vishalbhoj . If `build_status` is historical, then it sounds like we should stay on `result` for now. Thanks again. 👍

PR #156, opened 2021-06-29
Use ld.bfd for powerpc64le on -next with LLVM 11

nathanchance:


generator.yml
```diff
@@ -538,7 +538,8 @@ builds:
   # ppc32 and ppc64: Build disabled (https://gitlab.com/Linaro/tuxmake/-/issues/108)
   # - {<< : *ppc32,             << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_11}
   # - {<< : *ppc64,             << : *next,             << : *ppc64_llvm,      boot: true,  llvm_version: *llvm_11}
-  - {<< : *ppc64le,           << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_11}
+  # ppc64le: Build disabled (https://github.com/ClangBuiltLinux/linux/issues/1260)
+  # - {<< : *ppc64le,           << : *next,             << : *llvm,            boot: true,  llvm_version: *llvm_11}
```
nickdesaulniers, line 542: We do have a `ppc64_llvm` "tier" that enables BFD. Could we reuse that here to keep the build going?  That tier exists to work around https://github.com/ClangBuiltLinux/linux/issues/602 but we could add a link in a comment to https://github.com/ClangBuiltLinux/linux/issues/1260.

nathanchance: I believe so, let's test it :)

nickdesaulniers (approved):
thanks! 🚚 🚯

PR #158, opened 2021-06-30
Remove CROSS_COMPILE_COMPAT for upstream 5.4

nathanchance:
The compat vDSO for arm64 did not build upstream until 5.7 because of
the series associated with commit a5d442f50a41 ("arm64: vdso32: Enable
Clang Compilation"). That series was backported to Android 5.4 so we do
not need any changes there.

Link: https://github.com/ClangBuiltLinux/continuous-integration2/runs/2947999652?check_suite_focus=true


nickdesaulniers (approved):
thanks, I was just looking at this failure wondering what was going on. 🧑‍🍳 🌁

PR #159, opened 2021-06-30
Move RISC-V back to ld.bfd on 5.10

nathanchance:
5.10 has a different issue that was tangentially fixed in mainline,
requiring further investigation.


nickdesaulniers (approved):
Hmm...

Thanks for the patch though! 🚒 🍸

PR #171, opened 2021-07-24
Enable more configurations for ARCH=arm

nathanchance:


nickdesaulniers:
Do these fail to boot?

nathanchance:
I did not try. I figured `omap2plus_defconfig` and `imx_v4_v5_defconfig` probably wouldn't boot under the QEMU virt CPU and I forgot to test the LPAE config.

nathanchance:
Okay, I have enabled the testing for the LPAE config but the other two configs will not work with `boot-qemu.sh` in its current form since we boot specific v5 and v6 machines.

nathanchance:
I am going to merge this without explicit approval after our conversation offline. Please scream if you see anything wrong after the fact and I am more than happy to push subsequent fixups.

PR #172, opened 2021-07-24
Add some distribution configurations

nathanchance:
Arch, Fedora, and OpenSUSE to start because their configurations are easily accessible from the web.

This is currently blocked on https://gitlab.com/Linaro/tuxsuite/-/issues/124 in its current form, I am just pushing for early review.

### DIFF
--- a/generator.yml
+++ b/generator.yml
@@ -115,9 +115,9 @@ configs:
   - &ppc32             {config: [ppc44x_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                               kernel_image: uImage,       << : *powerpc-triple,       << : *kernel_modules}
   - &ppc64             {config: [pseries_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: vmlinux,      << : *powerpc64-triple,     << : *kernel_modules}
   - &ppc64le           {config: [powernv_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
-  # CONFIG_BPF_PRELOAD disabled for all cross compiled Fedora configs: https://github.com/ClangBuiltLinux/linux/issues/1433
-  - &ppc64le_fedora    {config: [*ppc64le-fedora-config-url, CONFIG_BPF_PRELOAD=n],                                                        << : *powerpc64le-triple,   << : *kernel_modules}
-  - &ppc64le_suse      {config: *ppc64le-suse-config-url,                                                                                  << : *powerpc64le-triple,   << : *kernel_modules}
+  # CONFIG_BPF_PRELOAD disabled for all cross compiled Fedora configs: https://github.com/ClangBuiltLinux/linux/issuen/1433
+  - &ppc64le_fedora    {config: [*ppc64le-fedora-config-url, CONFIG_BPF_PRELOAD=n],                            kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
+  - &ppc64le_suse      {config: *ppc64le-suse-config-url,                                                      kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
   - &riscv             {config: defconfig,                                                                     kernel_image: Image,        << : *riscv-triple,         << : *kernel_modules}
   #                                         https://github.com/ClangBuiltLinux/linux/issues/1143
   - &riscv_no_efi      {config: [defconfig, CONFIG_EFI=n],                                                     kernel_image: Image,        << : *riscv-triple,         << : *kernel_modules}
@@ -171,8 +171,8 @@ builds:
   - {<< : *arm32_allmod,      << : *mainline,         << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
   - {<< : *arm32_allno,       << : *mainline,         << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
   - {<< : *arm32_allyes,      << : *mainline,         << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
-  - {<< : *arm32_fedora,      << : *mainline,         << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
-  - {<< : *arm32_suse,        << : *mainline,         << : *llvm_full,       boot: false, llvm_version: *llvm_tot}
+  - {<< : *arm32_fedora,      << : *mainline,         << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
+  - {<< : *arm32_suse,        << : *mainline,         << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
   - {<< : *arm64,             << : *mainline,         << : *arm64_llvm_full, boot: true,  llvm_version: *llvm_tot}
   - {<< : *arm64be,           << : *mainline,         << : *llvm_full,       boot: true,  llvm_version: *llvm_tot}
   - {<< : *arm64_lto_full,    << : *mainline,         << : *arm64_llvm_full, boot: true,  llvm_version: *llvm_tot}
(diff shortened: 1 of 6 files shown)


nathanchance:
https://gitlab.com/Linaro/tuxsuite/-/issues/124 has been resolved so we can do this now.

Certain builds are broken in -next but that is just -next for ya :^)

generator.yml
```diff
@@ -63,63 +63,63 @@ targets:
...
-  - &i386              {config: defconfig,                                                                                                 << : *i386-triple,          << : *kernel_modules}
-  - &i686_fedora       {config: [https://src.fedoraproject.org/rpms/kernel/raw/rawhide/f/kernel-i686-fedora.config, CONFIG_BPF_PRELOAD=n], << : *i386-triple,          << : *kernel_modules}
-  - &i386_suse         {config: https://github.com/openSUSE/kernel-source/raw/master/config/i386/default,                                  << : *i386-triple,          << : *kernel_modules}
-  - &mips              {config: [malta_defconfig, CONFIG_BLK_DEV_INITRD=y, CONFIG_CPU_BIG_ENDIAN=y],           kernel_image: vmlinux,      << : *mips-triple,          << : *kernel_modules}
-  - &mipsel            {config: [malta_defconfig, CONFIG_BLK_DEV_INITRD=y],                                    kernel_image: vmlinux,      << : *mipsel-triple,        << : *kernel_modules}
+  - &arm64_suse        {config: [https://github.com/openSUSE/kernel-source/raw/master/config/arm64/default, CONFIG_DEBUG_INFO_BTF=n],         << : *arm64-triple,         << : *kernel_modules}
```
nickdesaulniers, line 90: rather than re-align because we have these huge URLs inline, can we add these URLs to the list of URLs at the top of this file, then reference them with a shorter identifier inline?

nathanchance: If I understand you correctly?

```yaml
urls:
...
- &arm64-suse-config-url: https://github.com/openSUSE/kernel-source/raw/master/config/arm64/default
...
configs:
...
  - &arm64_suse        {config: [*arm64-suse-config-url, CONFIG_DEBUG_INFO_BTF=n], << : *arm64-triple, << : *kernel_modules}
...
```

Sure, no worries, I will do that shortly.

nathanchance: Done. PTAL.

generator.yml
```diff
@@ -71,6 +86,8 @@ configs:
   - &arm32_allmod      {config: allmodconfig,                                                                                              << : *arm-triple,           << : *kernel_modules}
   - &arm32_allno       {config: allnoconfig,                                                                                               << : *arm-triple,           << : *kernel_modules}
   - &arm32_allyes      {config: allyesconfig,                                                                                              << : *arm-triple,           << : *kernel_modules}
+  - &arm32_fedora      {config: [*arm32-fedora-config-url, CONFIG_BPF_PRELOAD=n],                                                          << : *arm-triple,           << : *kernel_modules}
```
nickdesaulniers, line 89: can you add a comment with link to why BPF_PRELOAD is disabled?

nickdesaulniers: ah, it's below. add it here, too?

generator.yml
```diff
@@ -82,18 +99,28 @@ configs:
...
   # PowerPC builds disable -Werror for now: https://github.com/ClangBuiltLinux/linux/issues/1386
   - &ppc32             {config: [ppc44x_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                               kernel_image: uImage,       << : *powerpc-triple,       << : *kernel_modules}
   - &ppc64             {config: [pseries_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: vmlinux,      << : *powerpc64-triple,     << : *kernel_modules}
   - &ppc64le           {config: [powernv_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
+  - &ppc64le_fedora    {config: [*ppc64le-fedora-config-url, CONFIG_BPF_PRELOAD=n],                                                        << : *powerpc64le-triple,   << : *kernel_modules}
+  - &ppc64le_suse      {config: *ppc64le-suse-config-url,                                                                                  << : *powerpc64le-triple,   << : *kernel_modules}
```
nickdesaulniers, line 117: do we want to disable WERROR for these, too?

nickdesaulniers (commented):
can we boot test these too, please? Or don't they boot?

generator.yml
nathanchance: No, they build fine with `-Werror`. I think we can actually re-enable `-Werror` for the other configurations but that is tangential to this.

generator.yml
nathanchance: Sure.

nathanchance:
> can we boot test these too, please? Or don't they boot?

They should boot fine as I boot test these locally but I will have to modify the logic in `get_cbl_name()` to account for these. Let me think about how to do that.

nathanchance:
Alright, I believe this should work fine, as `get_cbl_name()` returns what I would expect it to.

utils.py
```diff
@@ -29,6 +29,35 @@ def get_cbl_name():
...
+                "aarch64": "arm64",
+                "armv7hl": "arm32_v7",
+                "i686": "x86",
+                "s390x": "s390"
+            }
+            # The URL is https://.../kernel-<arch>-fedora.config
+            fedora_arch = base_config.split("/")[-1].split("-")[1]
+            if fedora_arch in fedora_to_cbl:
+                return fedora_to_cbl[fedora_arch]
+            return fedora_arch
```
nickdesaulniers, line 45: perhaps just `return fedora_to_cbl[fedora_arch]` so that we fail loudly for unknown arches? Or are there values not in `fedora_to_cbl`?

utils.py
```diff
@@ -29,6 +29,35 @@ def get_cbl_name():
...
+            suse_to_cbl = {
+                "armv7hl": "arm32_v7",
+                "i386": "x86",
+                "s390x": "s390"
+            }
+            # The URL is https://.../<arch>/default
+            suse_arch = base_config.split("/")[-2]
+            if suse_arch in suse_to_cbl:
+                return suse_to_cbl[suse_arch]
+            return suse_arch
```
nickdesaulniers, line 56: perhaps just `return suse_to_cbl[suse_arch]` so that we fail loudly for unknown arches? Or are there values not in suse_to_cbl?

generator.yml
```diff
@@ -115,9 +115,9 @@ configs:
   - &ppc32             {config: [ppc44x_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                               kernel_image: uImage,       << : *powerpc-triple,       << : *kernel_modules}
   - &ppc64             {config: [pseries_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: vmlinux,      << : *powerpc64-triple,     << : *kernel_modules}
   - &ppc64le           {config: [powernv_defconfig, CONFIG_PPC_DISABLE_WERROR=y],                              kernel_image: zImage.epapr, << : *powerpc64le-triple,   << : *kernel_modules}
-  # CONFIG_BPF_PRELOAD disabled for all cross compiled Fedora configs: https://github.com/ClangBuiltLinux/linux/issues/1433
-  - &ppc64le_fedora    {config: [*ppc64le-fedora-config-url, CONFIG_BPF_PRELOAD=n],                                                        << : *powerpc64le-triple,   << : *kernel_modules}
-  - &ppc64le_suse      {config: *ppc64le-suse-config-url,                                                                                  << : *powerpc64le-triple,   << : *kernel_modules}
+  # CONFIG_BPF_PRELOAD disabled for all cross compiled Fedora configs: https://github.com/ClangBuiltLinux/linux/issuen/1433
```
nickdesaulniers, line 118: typo?

nickdesaulniers: s/issuen/issues/

utils.py
nathanchance: `ppc64le` and `x86_64` are left out of `fedora_to_cbl` because they are the same; I can add them and just `return fedora_to_cbl[fedora_arch]` if you prefer?

utils.py
nathanchance: `arm64`, `ppc64le`, and `x86_64` are left out because they are a one to one map. As above, I can add them and just `return suse_to_cbl[suse_arch]` if you prefer?

generator.yml
nathanchance: Done.

nathanchance:
I am going to merge this without explicit approval after our conversation offline. Please scream if you see anything wrong after the fact and I am more than happy to push subsequent fixups.

nathanchance:
I pushed https://github.com/ClangBuiltLinux/continuous-integration2/commit/f9f368d3e2364e774de44f83d3aa6563f2bee054 to make the build green. I am not sure it warrants an investigation on our part given it happens with GCC but I can open a follow up issue if you feel it is worth looking into.

PR #177, opened 2021-08-02
Re-enable -Werror for `ARCH=powerpc`

nathanchance:


nickdesaulniers:
hmm...I kind of prefer these builds to be less noisy.  It's not like we have -Werror set for other builds, so we're not catching warnings for other ARCHs unless we go look at the logs.

nathanchance:
`-Werror` only impacts `arch/powerpc` and it is how the defconfigs are upstream so if we ignore `-Werror` and someone else tries to go build a defconfig that has a warning in `arch/powerpc`, then the build is broken for them and not us, which is a poor experience to me. It costs us nothing to turn it off again if it breaks the build 🤷‍♂️

nickdesaulniers (approved):
easy enough to re-disable if needed 👍 🍷

Issue #176, opened 2021-07-28
Handle transition from LLVM 13 to LLVM 14


nathanchance:
The `release/13.x` branch has been cut meaning `clang-nightly` will now be `clang-14` and there will eventually be `clang-13` through [apt.llvm.org](https://apt.llvm.org/). The transition plan:

- [x] Wait for LLVM 13 binaries on [apt.llvm.org](https://apt.llvm.org/) ([done](https://lore.kernel.org/r/89b3f65d-b2c8-bd39-24a6-fd6b7f146a98@mozilla.com/))
- [x] Report availability of LLVM 13 binaries on [apt.llvm.org](https://apt.llvm.org/) to [TuxMake issue tracker](https://gitlab.com/Linaro/tuxmake) ([done](https://gitlab.com/Linaro/tuxmake/-/issues/154))
- [x] Wait for TuxMake to provide `clang-13` images ([done](https://gitlab.com/Linaro/tuxmake/-/commit/ffe969123598db11fbe196bd211385472f8485fe))
- [x] Wait for TuxSuite to allow `clang-13` to be specified ([done](https://gitlab.com/Linaro/tuxsuite/-/issues/129))
- [x] Update `generator.yml` to shift `llvm_tot` to 14 ([done](https://github.com/ClangBuiltLinux/continuous-integration2/pull/179))
- [x] Update `generator.yml` to shift `llvm_latest` to 13 and copy all `llvm_tot` builds to it ([done](https://github.com/ClangBuiltLinux/continuous-integration2/pull/179))
- [x] Update `generator.yml` to provide `llvm_12` anchor and copy all `llvm_latest` builds to it ([done](https://github.com/ClangBuiltLinux/continuous-integration2/pull/179))

cc @terceiro

terceiro:
ack. I'm watching this and will do the tuxmake side when appropriate

nathanchance:
41fc75f3c2d4a5439d6bb1a7088c6853ee9ae57d

PR #179, opened 2021-08-08
Update llvm_tot anchor and LLVM_TOT_VERSION file

nathanchance:
Part of #176.


generator.yml
```diff
@@ -1,5 +1,5 @@
 llvm_versions:
-  - &llvm_tot 13
+  - &llvm_tot 14
   - &llvm_latest 12
```
nickdesaulniers, line 3: Do you want to add anchors for `llvm_12` and bump `llvm_latest` to 13?

LLVM_TOT_VERSION
```diff
@@ -1 +1 @@
-13
+14
```
nickdesaulniers, line 1: Doesn't running `generate.sh` update this? Has it been being modified for previous patches, just not committed?

nickdesaulniers (approved):

LLVM_TOT_VERSION
nathanchance: Yes to both questions.

generator.yml
nathanchance: Not yet, TuxSuite does not have clang-13 yet: https://gitlab.com/Linaro/tuxsuite/-/issues/129

I can change the `llvm_latest` anchor to `llvm_12` and update all of the builds accordingly with this pull request then add `llvm_latest` back as `13` when the images are available? Choice is up to you.

nathanchance:
Looks like `clang-13` is live in TuxSuite so I will update this PR to do the rest of #176.

https://gitlab.com/Linaro/tuxsuite/-/issues/129

nathanchance:
Already, this should close out #176.

For the `clang-13` builds, I just copied all of the builds we are doing for LLVM 14. If you want me to slim up the list, I can but it might not be a terrible idea to keep all of that coverage so we can spot ToT breakages more easily.

PR #180, opened 2021-08-08
Add separate distribution configuration build set

nathanchance:
Right now, the distribution configurations are lumped in with the
all*configs set, which means that we cannot get boot results until all
of the all*configs are done building.

Separate out the distribution configs into their own set so that they
can be built then boot tested in parallel with the defconfigs and
allconfigs build sets.


nickdesaulniers (approved):

nickdesaulniers:
Does this imply we'll have new buttons we need to add for new branches?

nathanchance:
No, the buttons link to the workflow runs, not the individual steps within the workflows.

Thanks for the review!

PR #184, opened 2021-08-12
Error out if builds.json is not found

nathanchance:
When the tuxsuite CLI errors out because of invalid input (such as an
invalid make variable), the "tuxsuite" step in GitHub Actions passes
but all the check_logs.py steps error because builds.json is not found.
This is a little cryptic because it seems like the kernel builds
happened but they really didn't.

To make it a little easier to triage failures like this, make the upload
artifacts step error out if builds.json is not found, which means the
tuxsuite client errored out before builds were started or there was
something fatal with the builds that did happen. No builds.json file
means that check_logs.py cannot run at all so there is no point in even
letting those jobs run.

### DIFF
--- a/.github/workflows/4.14.yml
+++ b/.github/workflows/4.14.yml
@@ -31,6 +31,7 @@ jobs:
       with:
         path: builds.json
         name: output_artifact_defconfigs
+        if-no-files-found: error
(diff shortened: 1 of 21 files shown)


nickdesaulniers (approved):
awesome thank you! 🥝 🧇

PR #185, opened 2021-08-14
check_logs.py: Increase timeout for CONFIG_KASAN_SW_TAGS builds

nathanchance:
clang-11 and clang-12 builds with CONFIG_KASAN_SW_TAGS=y are incredibly
slow to boot up but clang-13 and newer builds are fine. Increase the
timeout if CONFIG_KASAN_SW_TAGS=y is in the list of configs so that we
are sure that the build timed out rather than being slow.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -79,7 +79,11 @@ def run_boot(build):
     if "CONFIG_KASAN=y" in build["kconfig"] or \
        "CONFIG_KCSAN=y" in build["kconfig"] or \
        "CONFIG_UBSAN=y" in build["kconfig"]:
-        boot_qemu += ["-s", "4", "-t", "10m"]
+        boot_qemu += ["-s", "4"]
+        if "CONFIG_KASAN=y" in build["kconfig"]:
+            boot_qemu += ["-t", "15m"]
+        else:
+            boot_qemu += ["-t", "10m"]
     try:
         subprocess.run(boot_qemu, check=True)
     except subprocess.CalledProcessError as e:


check_logs.py
```diff
@@ -79,7 +79,11 @@ def run_boot(build):
     if "CONFIG_KASAN=y" in build["kconfig"] or \
        "CONFIG_KCSAN=y" in build["kconfig"] or \
        "CONFIG_UBSAN=y" in build["kconfig"]:
-        boot_qemu += ["-s", "4", "-t", "10m"]
+        boot_qemu += ["-s", "4"]
+        if "CONFIG_KASAN_SW_TAGS=y" in build["kconfig"]:
+            boot_qemu += ["-t", "30m"]
```
nickdesaulniers, line 84: Honestly, if these are taking >10m to boot, I question how reasonable it is for us to even have test coverage of these configs.

nathanchance: That is just the nature of GitHub Actions unfortunately, it is a two core virtual machine so we paying a performance penalty versus running on bare metal on top of the emulation penalty. Booting these locally takes considerably less time. I can disable the boot tests if you prefer.

nickdesaulniers: @ramosian-glider @eugenis @pcc @melver what's the value of maintaining `CONFIG_KASAN_SW_TAGS`?

ojeda: Another approach would be to see if Google etc. are willing to lend us some beefy VMs for ClangBuiltLinux / Rust for Linux.

melver: > @ramosian-glider @eugenis @pcc @melver what's the value of maintaining `CONFIG_KASAN_SW_TAGS`?
@xairy 

SW_TAGS should still be tested if possible.

However, it might be fair to limit its _boot_ testing to latest Clang if that helps. If possible, still compile-test other versions but just not boot them.

ivoire: Sorry to come by, but why don't you use TuxTest for the boot (and ltp tests) like you use TuxBuild for the builds?

nathanchance: > SW_TAGS should still be tested if possible.
> 
> However, it might be fair to limit its _boot_ testing to latest Clang if that helps. If possible, still compile-test other versions but just not boot them.

There appears to be some flakiness around how long the boots take on GitHub Actions, sometimes even LLVM 14 fails to boot within 10 minutes:

https://github.com/ClangBuiltLinux/continuous-integration2/runs/3465520254?check_suite_focus=true

I can either disable the boot tests altogether or try experimentally increasing the timeout length in intervals of 5 minutes so that we are not taking a ton of time on the boot. It seems like I might need to do this for `CONFIG_KASAN=y` in general as there was a `CONFIG_KASAN_VMALLOC=y` build that timed out:

https://github.com/ClangBuiltLinux/continuous-integration2/runs/3465520199?check_suite_focus=true

To be entirely fair, it is not like we are running presubmit build testing so does the time that the machine takes to load really actually matter? We could increase the timeout length to an hour and I do not think it would really impact us because we are running on GitHub Actions machines, not TuxSuite.

nickdesaulniers: I guess it's a question of "at what point will github actions go travisci on us and require us to pay for time?" If we can get away with it, go for it, but let's do 5 minute increments vs whole hours.

nathanchance: > I guess it's a question of "at what point will github actions go travisci on us and require us to pay for time?"

At that point, we can re-evaluate how much time we are taking and disable boots that are taking too long.

[Their usage limits](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits) tend to be pretty generous in terms of time, although I see one limit that we are kind of pushing up against, which I will open an issue for shortly.

> If we can get away with it, go for it, but let's do 5 minute increments vs whole hours.

Sure, I have increased it to 15m, PTAL. If we keep getting failures, I'll increase it to 20m and so on and so forth.

nickdesaulniers (approved):

PR #178, opened 2021-08-08
Enable sanitizer builds

nathanchance:
Closes: #169

PR #187, opened 2021-08-31
Test CONFIG_KASAN on s390 with clang-13+

nathanchance:
We enabled support for testing `CONFIG_KASAN` in #178 but I did not realize that s390 supported KASAN so let's add it for more coverage. This passes my local tests.

32-bit ARM also supports `CONFIG_KASAN` but unfortunately, there is [an outstanding bug](https://github.com/ClangBuiltLinux/linux/issues/1211), which is resolved with @nickdesaulniers 's [patch](https://lore.kernel.org/r/20210830213846.2609349-1-ndesaulniers@google.com/). Once that patch has been merged, we can do this same thing for `ARCH=arm`.

### DIFF
--- a/.github/workflows/mainline.yml
+++ b/.github/workflows/mainline.yml
@@ -536,6 +536,27 @@ jobs:
       run: echo "::add-matcher::.github/problem-matchers/clang-errors-warnings.json"
     - name: Boot Test
       run: ./check_logs.py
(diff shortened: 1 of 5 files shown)


nickdesaulniers (approved):
thanks! 🥥 ⛳

nathanchance:
Thanks for the review, let's see what breaks :)

Issue #189, opened 2021-09-01
disable -Werror for ppc again


nickdesaulniers:
https://github.com/ClangBuiltLinux/continuous-integration2/runs/3484727335?check_suite_focus=true
until https://github.com/ClangBuiltLinux/linux/issues/1445 is resolved

PR #190, opened 2021-09-02
Disable -Werror for pseries_defconfig for now

nathanchance:


nickdesaulniers (approved):
consider the use of `Fixes` rather than `Link` in the commit message to auto-close #189 when PR is merged. 🐺 🌫️

nathanchance:
Yes, realized after the fact and just linked this PR against #189 :)

Thanks for the review!

PR #191, opened 2021-09-02
Check matrices for number of jobs

nathanchance:
GitHub Actions has a maximum of 256 jobs per matrix. In order to make
sure that we stay under this, add a workflow to check the number of jobs
in a TuxSuite matrix. The other workflows will never hit this number of
jobs so we just ignore them.

### DIFF
--- a/check-matrix.sh
+++ b/check-matrix.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [[ -n ${GITHUB_ACTIONS} ]]; then
+    repo=${GITHUB_WORKSPACE}
+else
+    repo=$(dirname "$(readlink -f "${0}")")
+fi
+
+for workflow in "${repo}"/.github/workflows/*.yml; do
+    # Get number of jobs from workflow (kick_tuxsuite... and md5 names)
+    jobs=$(grep -cP "  _|  kick" "${workflow}")
+    workflow=$(basename "${workflow}")
+
+    # If we did not find any jobs in the following format, it is not a TuxSuite
+    # workflow and we don't care.
+    [[ ${jobs} -eq 0 ]] && continue
+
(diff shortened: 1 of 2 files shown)


check-matrix.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+for workflow in "${GITHUB_WORKSPACE}"/.github/workflows/*.yml; do
```
nickdesaulniers, line 3: When I run this locally, I get:
```
grep: /.github/workflows/*.yml: No such file or directory
```
because `GITHUB_WORKSPACE` is undefined in my local env, and so we have a preceding slash. Can we make this work locally, too?

nathanchance: Ah good catch, will fix.

nathanchance: Done, PTAL.

check-matrix.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [[ -n ${GITHUB_ACTIONS} ]]; then
+    repo=${GITHUB_WORKSPACE}
+else
+    repo=$(dirname "$(readlink -f "${0}")")
```
nickdesaulniers, line 6: what's wrong with good ole `pwd`?

nickdesaulniers (approved):

check-matrix.sh
nathanchance: I write all of my `bash` scripts under the assumption that they should work from wherever they are being called (i.e. `../ci/check-matrix.sh` does not work). I can change this to `${PWD}` if you prefer, given that it is unlikely that the script will be called from somewhere other than the repo.

nickdesaulniers: cool, no worries

PR #194, opened 2021-09-08
Disable -Werror for now

nathanchance:
`CONFIG_WERROR` is enabled by default when `CONFIG_COMPILE_TEST` is enabled.
Disable it explicitly so we can catch actual build breakage.

I am of the opinion we should prioritize getting warning clean so that
we can re-enable this, as this does not affect our boot tests anymore
but will make it obvious when there are potential issues due to
warnings.


nickdesaulniers (approved):
I agree. Thanks for this patch. 🈂️ 🔧

nathanchance:
Thanks for the review!

PR #199, opened 2021-09-17
Add .config artifact checker

kees:
I wanted to learn the CI system, so I ran off and implemented a solution to the feature request I reported. ;)

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -101,6 +101,14 @@ def get_requested_llvm_version():
     return "clang-" + ("nightly" if ver == llvm_tot_version else ver)
 
 
+def show_builds():
+    print_yellow("Available builds:")
+    for build in _read_builds():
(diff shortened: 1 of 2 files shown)


nathanchance (approved):
This looks great to me! @nickdesaulniers any comments?

utils.py
```diff
@@ -81,11 +81,14 @@ def get_cbl_name():
 
 
 def _read_builds():
+    builds = "builds.json"
+    if os.environ.get("MOCK"):
```
nickdesaulniers, line 85: We should document the use of this env var in the README.

nickdesaulniers (approved):
great, thank you! 🛷 🌵

roxell:
This seems like a good feature, would it be good to push this into the kernel itself?

PR #201, opened 2021-09-18
Flush output buffers proactively

kees:
In the CI logs, QEMU output shows up before the rest of check_log.py
output. This is likely due to IO buffering, so explicitly flush buffers
in the helper routines and before spawning QEMU.

Signed-off-by: Kees Cook <keescook@google.com>

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -126,10 +126,12 @@ def get_build():
 
 def print_red(msg):
     print("\033[91m%s\033[0m" % msg, file=sys.stderr)
+    sys.stderr.flush()
 
 
 def print_yellow(msg):
-    print("\033[93m%s\033[0m" % msg)
+    print("\033[93m%s\033[0m" % msg, file=sys.stdout)
+    sys.stdout.flush()
 
 
 def patch_series_flag(tree):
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -113,7 +113,9 @@ def get_build():
 
 def print_red(msg):
     print("\033[91m%s\033[0m" % msg, file=sys.stderr)
+    sys.stderr.flush()
```
MaskRay, line 116: stderr is unbuffered and many applications leverage this property. No need for flush.

nickdesaulniers: Indeed, see also `man 3 stderr`:

> The  stream  stderr  is  unbuffered.

nickdesaulniers (changes requested):

utils.py
nickdesaulniers: Though note that it seems that @kees is modifying the buffering of stderr for our boot wrapper script in https://github.com/ClangBuiltLinux/boot-utils/pull/49.

kees:
What changes are desired here? It's not harmful to flush stderr. At best it fixes intermixed output, and at worst it is redundant. :)

nathanchance (approved):

nickdesaulniers:
Sure, but at the least the branch needs to be rebased on main and force pushed before we can merge.

kees:
Okay, rebased. :)

nickdesaulniers (approved):
doesn't matter to me

PR #204, opened 2021-09-21
Disable boot for CONFIG_KASAN_SW_TAGS=y on clang-11 and clang-12

nathanchance:
After increasing the timeout to 20m, there are still boot timeouts for
clang-11 and clang-12. It is likely that we would need to increase the
timeout to 40+ minutes so instead, just disable the boot tests.


nickdesaulniers (approved):
do we want to bring the timeout back down then? Thanks for the patch! 🧥 🇦🇨

nathanchance:
I would rather leave it as is because there are times where even the newer clang versions have taken up to 15 minutes. Better to be safe than sorry. Thanks for the review!

PR #208, opened 2021-09-22
check_logs.py: Handle equals sign in configuration value

nathanchance:
The current code chokes when there is an equals sign in the
configuration value:

`CONFIG_CMDLINE="console=ttyAMA0"`

```
  File "./check_logs.py", line 169, in <module>
    check_built_config(build)
  File "./check_logs.py", line 85, in check_built_config
    name, state = line.split('=')
ValueError: too many values to unpack (expected 2)
```

We just want to split on the first equals sign so add a max split of 1.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -82,7 +82,7 @@ def check_built_config(build):
         name = None
         state = None
         if '=' in line:
-            name, state = line.split('=')
+            name, state = line.split('=', 1)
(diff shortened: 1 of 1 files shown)


nickdesaulniers (approved):
ah good catch; thanks! 🟧 🟪

kees:
Eek! Thanks for fixing that!

PR #210, opened 2021-09-22
patches: Update FPU patch to upstream submission

nathanchance:
While the actual content of the patch is the same, it provides a link
to see what the submission/acceptance status is, which helps us move
patches along if they get stuck.


nickdesaulniers (approved):
sgtm thanks! 🎲 🧣

PR #211, opened 2021-09-22
problem matcher updates

kees:
This add problem matching for more tool outputs and kernel Oopses.


nathanchance:
A couple of comments:

1. Do we need to disable the oops check for the KCSAN KUnit tests as well?

2. I am not really sure that I care for the `dtc` warning highlighting. I know the build should be as clean as possible but I would think that `dtc` warnings are outside of the scope of the project, given that the compiler has nothing to do with them? I do not have a strong opinion though :)

kees:
> A couple of comments:
> 
> 1. Do we need to disable the oops check for the KCSAN KUnit tests as well?

Oh! Yes, good call. My eyes missed the "C" difference from KASAN. :)

> 2. I am not really sure that I care for the `dtc` warning highlighting. I know the build should be as clean as possible but I would think that `dtc` warnings are outside of the scope of the project, given that the compiler has nothing to do with them? I do not have a strong opinion though :)

Hmm. You make a good point about "scope of the project". Those are warnings, but they genuinely have nothing to do with clang. Okay, I'll drop that.

nathanchance:
Taking this for a spin on my own repo, it looks like there appears to be an error with one of the matchers: https://github.com/nathanchance/continuous-integration2/runs/3681335745?check_suite_focus=true

I think this is complaining about 43ed085ae2d23dec34a21ae8bd8961282c54ea6d? I am not sure, the error is kind of cryptic but it makes the most sense. It is possible that should be split off like the oops one was.

kees:
Maybe it _isn't_ legal to have multiple matchers? Hmmm

nathanchance:
The problem matcher examples show multiple matchers in different json files, perhaps that would work?

nathanchance:
This one looks to be good!

https://github.com/nathanchance/continuous-integration2/actions/runs/1267235942

nathanchance (approved):

generate_workflow.py
```diff
@@ -158,11 +158,7 @@ def get_steps(build, build_set):
                     },
                 },
                 {
-                    "name": "Register clang error/warning problem matcher",
-                    "run": 'echo "::add-matcher::.github/problem-matchers/clang-errors-warnings.json"'
-                },
-                {
-                    "name": "Boot Test",
```
nickdesaulniers, line 165: Is it easy to break this into 2 commits; one for the python script changes, one for regen'ing the workflow files?

kees: I can do that, sure. It seemed like it should go together to avoid having a state where the problem matcher got registered twice.

nickdesaulniers (approved):
looks great! thanks for all of the work that went into these.  Maybe @mpe might be interested in some of these, since ppc is also testing via github actions?

PR #212, opened 2021-09-23
patches: Remove FPU patch for -next and -tip

nathanchance:
It has been merged into -tip and auto-latest has been regenerated,
meaning it will be in the -next release tomorrow.


nickdesaulniers (approved):
do you want to remove it from lto-cfi-tip, too?

nathanchance:
@samitolvanen will need to update that branch before we can. Hint hint? :)

Issue #213, opened 2021-09-24
retry logic


nickdesaulniers:
we need some retry logic to not fail the build when we fail to fetch logs or images. examples:
https://github.com/ClangBuiltLinux/continuous-integration2/runs/3700664838?check_suite_focus=true
```
raise URLError(err)
urllib.error.URLError: <urlopen error [Errno 110] Connection timed out>
```
https://github.com/ClangBuiltLinux/continuous-integration2/runs/3700343954?check_suite_focus=true
```
ConnectionResetError: [Errno 104] Connection reset by peer
```
probably should implement expenontial backoff, too, and an upper threshold.

nathanchance:
So I assume this would look something like actually checking the URLError number and retrying if it were 104 or 110, up to five times or some number that we determine?

Tangentially, these errors seem more numerous now, is this a TuxSuite issue?

kees:
Maybe replace with `wget --tries=N` ?

vishalbhoj:
I have been analyzing the infrastructure if there is anything from TuxSuite end that we could do to improve this. After tracing the failure, I found out that 5xx errors happening from AWS storage service. The recommendation for such failures is to do an exponential backoff and retry mechanism[1]. We also use the same in tuxsuite cli while posting any requests to the API.

[1] https://aws.amazon.com/premiumsupport/knowledge-center/http-5xx-errors-s3/

PR #214, opened 2021-09-27
check_logs.py: Try multiple times to fetch items with exponential backoff

nathanchance:
There are times where there are issues with fetching items from
TuxSuite, usually as a result of Amazon Web Services timing out or
having an internal server error. These failures make it difficult to
tell when there is actually an issue worth looking into vs. "oh, just
another fetch failure".

To fix this, attempt to fetch the item at least seven times, backing off
by 2^(current retry number) seconds each time.

Certain errors, such as a 404 error, will result in an immediate
failure because there is nothing to fetch.

NOTE: I have not actually tested this but I think the logic is sound.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/213

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -4,6 +4,7 @@ import json
 import os
 import subprocess
 import sys
+import time
 import urllib.request
 
 from utils import get_build, get_image_name, print_red, print_yellow, get_cbl_name, show_builds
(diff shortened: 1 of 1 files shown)


check_logs.py
```diff
@@ -13,7 +14,36 @@
...
     # TODO: use something more robust like python wget library.
-    urllib.request.urlretrieve(url, dest)
+    retries = 0
+    max_retries = 7
+    retry_codes = [500, 504]
+    while retries < max_retries:
+        try:
+            time.sleep(2**retries)
+            urllib.request.urlretrieve(url, dest)
+            retries += 1
```
kees, line 24: retries += 1 needs to be before the urllib.request otherwise it'll never increment on failures.

check_logs.py
```diff
@@ -13,7 +14,36 @@
...
+                print_red(
+                    "%s could not be found (404 error), did the build timeout?"
+                    % (title))
+                sys.exit(1)
+            else:
+                print_red("%d error trying to download %s" % (err.code, title))
+                sys.exit(1)
+        except URLError:
+            pass
+
```
kees, line 41: I think it might be nice to include details about each failure case. Right now, this will silently retry with no logging of which of the conditional failures were encountered. i.e. perhaps something like this before each "pass" case:

```
   print_yellow("%s download failure: %s" % (str(err)))
   pass
```

kees (changes requested):

check_logs.py
```diff
@@ -13,7 +14,42 @@
 def _fetch(title, url, dest):
     print_yellow("fetching %s from: %s" % (title, url))
     # TODO: use something more robust like python wget library.
-    urllib.request.urlretrieve(url, dest)
+    retries = 0
+    max_retries = 7
+    retry_codes = [500, 504]
+    while retries < max_retries:
+        try:
+            time.sleep(2**retries)
```
nickdesaulniers, line 22:
```diff
- time.sleep(2**retries)
+ if (retries):
+   time.sleep(2 ** retries)
```
so that we don't sleep on the initial try.

nickdesaulniers (changes requested):

nickdesaulniers (approved):
🆒 thanks!

check_logs.py
```diff
@@ -13,7 +14,43 @@
...
+            if retries:
+                time.sleep(2**retries)
+            retries += 1
+            urllib.request.urlretrieve(url, dest)
+            break
+        except ConnectionResetError as err:
+            print_yellow('%s download error ("%s"), retrying...' %
+                         (title, str(err)))
+            pass
+        except HTTPError as err:
```
nickdesaulniers, line 31: I guess this isn't the right way to refer to this class of error in python: https://github.com/ClangBuiltLinux/continuous-integration2/runs/3767136408?check_suite_focus=true

> NameError: name 'HTTPError' is not defined

I think this needs to be `urllib.error.HTTPError`.

PR #215, opened 2021-09-28
Add problem matchers for depmod, dtc, and modpost

nathanchance:
The commit messages should sufficiently explain everything, let me know if there are any issues!

### DIFF
--- a/.github/problem-matchers/dtc.json
+++ b/.github/problem-matchers/dtc.json
@@ -1,0 +1,19 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match dtc parsing errors",
+            "owner": "dtc",
+            "pattern": [
+                {
+                    "__comment_regexp1": "Error: arch/arm64/boot/dts/qcom/sc7280-idp.dts:27.1-9 Label or path apps_rs not found",
+                    "regexp": "^(Error):\\s+([^:]+):([0-9]+)\\.([0-9]+)(?:[-0-9\\.]+)?\\s+(.*)$",
+                    "severity": 1,
+                    "file": 2,
+                    "line": 3,
+                    "column": 4,
+                    "message": 5
+                }
+            ]
+        }
+    ]
+}
--- a/.github/problem-matchers/generic.json
+++ b/.github/problem-matchers/generic.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match generic tool warnings/errors",
+            "owner": "generic",
+            "pattern": [
+                {
+                    "__comment_regexp1": "depmod: ERROR: Cycle detected: crypto -> crypto_algapi -> crypto",
+                    "__comment_regexp2": "depmod: ERROR: Found 2 modules in dependency cycles!",
+                    "regexp": "(^[^:]+:\\s+([^:]+):\\s+.*)$",
+                    "message": 1,
+                    "severity": 2
+                }
+            ]
+        }
+    ]
+}
--- a/.github/problem-matchers/modpost.json
+++ b/.github/problem-matchers/modpost.json
@@ -1,0 +1,16 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match modpost warnings/errors",
+            "owner": "modpost",
+            "pattern": [
+                {
+                    "__comment_regexp1": "ERROR: modpost: devm_drm_of_get_bridge [drivers/gpu/drm/vc4/vc4.ko] undefined!",
+                    "regexp": "^(WARNING|ERROR|FATAL):\\s+modpost:\\s+(.*)$",
+                    "severity": 1,
+                    "message": 2
+                }
+            ]
+        }
+    ]
+}
--- a/check_logs.py
+++ b/check_logs.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import glob
 import json
 import os
 import subprocess
@@ -122,7 +123,6 @@ def run_boot(build):
     ]
     if cbl_arch == "s390":
         boot_qemu += ["--use-cbl-qemu"]
-    problem_matcher = "linux-kernel-oopses.json"
     # If we are running a sanitizer build, we should increase the number of
     # cores and timeout because booting is much slower
     if "CONFIG_KASAN=y" in build["kconfig"] or \
@@ -136,12 +136,8 @@ def run_boot(build):
         if "CONFIG_KASAN_KUNIT_TEST=y" in build["kconfig"] or \
            "CONFIG_KCSAN_KUNIT_TEST=y" in build["kconfig"]:
             print_yellow(
-                "Skipping Oops problem matcher under Sanitizer KUnit build")
-            problem_matcher = None
-
-    if problem_matcher:
-        print_yellow("Register kernel Oops problem matcher")
-        print("::add-matcher::.github/problem-matchers/%s" % (problem_matcher))
+                "Disabling Oops problem matcher under Sanitizer KUnit build")
+            print("::remove-matcher owner=linux-kernel-oopses")
 
     # Before spawning a process with potentially different IO buffering,
     # flush the existing buffers so output is ordered correctly.
@@ -182,8 +178,8 @@ if __name__ == "__main__":
     build = get_build()
     print(json.dumps(build, indent=4))
     print_yellow("Register clang error/warning problem matchers")
-    print("::add-matcher::.github/problem-matchers/compiler-source.json")
-    print("::add-matcher::.github/problem-matchers/compiler-non-source.json")
+    for problem_matcher in glob.glob(".github/problem-matchers/*.json"):
+        print("::add-matcher::%s" % (problem_matcher))
     check_log(build)
     check_built_config(build)
     boot_test(build)


kees:
So, thinking a bit more about the problem matchers, I wonder if _any_ output during "make -s" should be considered a warning. :P That would need a lot more plumbing, though. Hmm.

.github/problem-matchers/depmod.json
```diff
@@ -0,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match depmod warnings/errors",
+            "owner": "depmod",
+            "pattern": [
+                {
+                    "__comment_regexp1": "depmod: ERROR: Cycle detected: crypto -> crypto_algapi -> crypto",
+                    "__comment_regexp2": "depmod: ERROR: Found 2 modules in dependency cycles!",
+                    "regexp": "^depmod:\\s+(FATAL|ERROR|WARNING|NOTICE|INFO|DEBUG):\\s+(.*)$",
```
kees, line 10: I though "severity" only knows about the literal strings "error" and "warning"? (Does this actually work for the other strings?)

.github/problem-matchers/dtc.json
```diff
@@ -0,0 +1,19 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match dtc parsing errors",
+            "owner": "dtc",
+            "pattern": [
+                {
+                    "__comment_regexp1": "Error: arch/arm64/boot/dts/qcom/sc7280-idp.dts:27.1-9 Label or path apps_rs not found",
+                    "regexp": "^(Error):\\s+([a-z|A-Z|0-9|\\/|.|\\-|_\\+]+):([0-9|\\-]+).([0-9|\\-]+)\\s+(.*)$",
```
kees, line 9: When I wrote the DTS error parser before, I noted its very unusual line/col output style. I don't think what you have here is correct. I think it spits out ${start_line}.${start_col}-[${end_line}.]${end_col}. So, I think you want:

...`:([0-9]+)\\.([0-9]+)-[0-9]+(?:\\.[0-9]+)\\s+(.*)$"`

(Note that the "." matches need to be `\\.` if you want to match a literal ".")

.github/problem-matchers/modpost.json
```diff
@@ -0,0 +1,16 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match modpost warnings/errors",
+            "owner": "modpost",
+            "pattern": [
+                {
+                    "__comment_regexp1": "ERROR: modpost: devm_drm_of_get_bridge [drivers/gpu/drm/vc4/vc4.ko] undefined!",
```
kees, line 8: Hah. Someone should send a patch to reverse the word order here. :) It really should be: `modpost: ERROR: ...` etc to match basically all the other tools. :)

.github/problem-matchers/depmod.json
kees, line 10: Also, given how roughly generic this could be if `^depmod:` were `^[^:]+:` should there be a catch-all problem matcher for "unknown" tools that follow this (common) convention?

.github/problem-matchers/modpost.json
```diff
@@ -0,0 +1,16 @@
...
+                    "regexp": "^(WARNING|ERROR|FATAL):\\s+modpost:\\s+(.*)$",
```
kees, line 9: Again, is "FATAL" actually handled sanely by Actions?

check_logs.py
```diff
@@ -182,8 +178,9 @@ def boot_test(build):
     build = get_build()
     print(json.dumps(build, indent=4))
     print_yellow("Register clang error/warning problem matchers")
-    print("::add-matcher::.github/problem-matchers/compiler-source.json")
-    print("::add-matcher::.github/problem-matchers/compiler-non-source.json")
+    problem_matchers = glob.glob(".github/problem-matchers/*.json")
+    for problem_matcher in problem_matchers:
```
kees, line 182: Oooh yeah, I like this a lot. :)

You could lose the "problem_matchers" intermediate variable if you wanted to, and just put the glob on the `for` line.

kees (changes requested):

.github/problem-matchers/depmod.json
nathanchance: Well, that is what the GitHub Actions documents say BUT the [setup-node](https://github.com/actions/setup-node/tree/main/.github) matchers also match on `info`... so I have no idea? Worst case, we roll the dice with this, it doesn't work, then we change the matching.

.github/problem-matchers/dtc.json
nathanchance: > I think it spits out ${start_line}.${start_col}-[${end_line}.]${end_col}.

Almost. It is either

* `${start_line}.${start_col}`
* `${start_line}.${start_col}-${end_col}`
* `${start_line}.${start_col}-${end_line}.${end_col}`

if I am reading the [dtc sources](https://git.kernel.org/pub/scm/utils/dtc/dtc.git/tree/srcpos.c#n296) correctly.

My current pattern should work for matching those BUT it does not getting the grouping correct, which appears hard to do because of how different the third form is from the first two. I think it is probably just work removing the grouping altogether and changing it to

`...:[0-9|\\.|\\-]+\\s+(.*)$`

Thoughts?

.github/problem-matchers/depmod.json
nathanchance: Yes, I think that is worthwhile so I will go ahead and change `depmod.json` to `generic.json`.

.github/problem-matchers/modpost.json
nathanchance: Hopefully :)

check_logs.py
nathanchance: Good call, I did that initially then I think I was going to do something with `problem_matchers` then never did.

.github/problem-matchers/depmod.json
kees: hmm! interesting. If it's an arbitrary string, we might be able to do really weird stuff with it. :) I wonder how it colorizes non-warn and non-error? I guess we'll find out. ;)

.github/problem-matchers/dtc.json
kees: I think this would work? `"^(Error):\\s+([^:]+):([0-9]+)\\.([0-9]+)(?:[-0-9\\.]+)?\s+(.*)$"`

.github/problem-matchers/generic.json
```diff
@@ -1,13 +1,13 @@
...
-            "__comment_owner": "match depmod warnings/errors",
-            "owner": "depmod",
+            "__comment_owner": "match generic tool warnings/errors",
+            "owner": "generic",
             "pattern": [
                 {
                     "__comment_regexp1": "depmod: ERROR: Cycle detected: crypto -> crypto_algapi -> crypto",
                     "__comment_regexp2": "depmod: ERROR: Found 2 modules in dependency cycles!",
-                    "regexp": "^depmod:\\s+(FATAL|ERROR|WARNING|NOTICE|INFO|DEBUG):\\s+(.*)$",
+                    "regexp": "^[^:]+:\\s+(FATAL|ERROR|WARNING|NOTICE|INFO|DEBUG):\\s+(.*)$",
```
kees, line 10: If this doesn't need to be strictly error/warning, how about:

`"^[^:]+:\\s+([^:]+):\\s+(.*)$"`

Though to capture the tool name, I wonder about:

`"(^[^:]+:\\s+([^:]+):\\s+.*)$"`

where message:1 and severity:2 :P

kees:
I'd say collapse the fixes into their respective commits and force-push the series when you've got the regexes fixed.

.github/problem-matchers/generic.json
nathanchance: I like the second one better because we show the full message, which will make triaging easily :)

.github/problem-matchers/dtc.json
nathanchance: Yes, that gets us close enough. I don't really think that we care that much about `dtc` errors given that they are never as a result of clang.

nathanchance:
@kees I gave you co-author credit on the first patch since I ended up using most of your regexs, hopefully you don't mind :)

kees:
> @kees I gave you co-author credit on the first patch since I ended up using most of your regexs, hopefully you don't mind :)

Damn this collaborative environment!  (Heh, yup, works for me.)

.github/problem-matchers/dtc.json
```diff
@@ -0,0 +1,19 @@
+{
+    "problemMatcher": [
+        {
+            "__comment_owner": "match dtc parsing errors",
+            "owner": "dtc",
+            "pattern": [
+                {
+                    "__comment_regexp1": "Error: arch/arm64/boot/dts/qcom/sc7280-idp.dts:27.1-9 Label or path apps_rs not found",
+                    "regexp": "^(Error):\\s+([^:]+):([0-9]+)\\.([0-9]+)(?:[-0-9\\.]+)?\s+(.*)$",
```
kees, line 9: There's a bare "\s" that should be "\\s". I'm impressed that the github UI showed this -- I would have totally missed this. :)

kees (changes requested):
Just one typo, otherwise looks great!

.github/problem-matchers/dtc.json
nathanchance: That is certainly impressive! I thought I escaped them all but it does not look like it... Should be fixed now!

kees (approved):

PR #216, opened 2021-09-30
Remove lto-cfi-tip patches

nathanchance:
The tip/clang-cfi branch has been updated to the latest tip/master,
which includes the patch we are carrying.


nickdesaulniers (approved):
thanks! 👔 🗝️

PR #217, opened 2021-10-01
fully qualify error classes from urllib

nickdesaulniers:
Otherwise we're referring to classes not in scope, as seen in:
https://github.com/ClangBuiltLinux/continuous-integration2/runs/3767136408?check_suite_focus=true

Link: #213
Link: #214
Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>


nathanchance (approved):
Thanks, sorry for not getting to this sooner (very long night...)

PR #218, opened 2021-10-01
patches: Drop RDMA patch from mainline

nathanchance:
It has been merged and will cause the entire run to fail.


nickdesaulniers:
this patch is also being applied to tip; should it be dropped from there, too?

nathanchance:
> this patch is also being applied to tip; should it be dropped from there, too?

Not until tip updates their master branch to include that patch.

nickdesaulniers (approved):
thanks, beat me to the punch! 🧑‍💼 🥩

PR #224, opened 2021-10-20
Fetch `metadata.json` and use it to display clang checkout date and hash

nathanchance:
Each commit message should hopefully be self explanatory, let me know if there are any issues or improvements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 dtbs/
 __pycache__/
 builds.json
+metadata.json
 *Image*
 vmlinux
--- a/check_logs.py
+++ b/check_logs.py
@@ -8,7 +8,7 @@ import sys
 import time
 import urllib.request
 
-from utils import get_build, get_image_name, print_red, print_yellow, get_cbl_name, show_builds
+from utils import get_build, get_image_name, get_requested_llvm_version, print_red, print_yellow, get_cbl_name, show_builds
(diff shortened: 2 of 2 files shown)


parse-debian-clang.sh
```diff
@@ -0,0 +1,51 @@
...
+    # $ clang-9 --version
+    # clang version 9.0.0-svn356030-1~exp1+0~20190313082415.405~1.gbp505100 (trunk)
+    # Target: x86_64-pc-linux-gnu
+    # Thread model: posix
+    # InstalledDir: /usr/bin
+    clang_date=$(clang --version | head -n1 | cut -d '~' -f 3 | cut -d . -f 1)
+
+    # Next, we need to parse the date into a format the date binary can understand
+    # We use bash substring expansion: https://wiki.bash-hackers.org/syntax/pe#substring_expansion
+    clang_date="${clang_date:0:4}-${clang_date:4:2}-${clang_date:6:2} ${clang_date:8:2}:${clang_date:10:2}:${clang_date:12:2}"
```
kees, line 41: I worry this is going to be fragile, but since it should fail loudly (i.e. clang_date will never be blank), I think it's okay.

nickdesaulniers (approved):
LGTM, thanks for all the work here. Let me add @sylvestre to reviewers to check those `cut` pipeline. 🅱️ 🐺

PR #230, opened 2021-11-02
Update mainline builds and patches

nathanchance:
Each commit should be fairly self explanatory.

NOTE: This does not get mainline back to green. It seems that there is a `BUILD_BUG_ON` failure with `clang-10` in `drivers/net/phy/marvell10g.c` that I did not have time to look into today. I really think we should consider bumping the minimum supported version of LLVM that we support in the kernel but that is a conversation for a different time.


kees (approved):
LGTM

nathanchance:
Thanks Kees!

PR #231, opened 2021-11-03
Fix more errors in next and mainline

nathanchance:
Each commit message should be self explanatory, let me know if there are any issues!


generator.yml
```diff
@@ -962,7 +956,7 @@ builds:
   - {<< : *ppc64le_suse,      << : *next,             << : *clang,           boot: true,  llvm_version: *llvm_11}
   # https://github.com/ClangBuiltLinux/linux/issues/1023
   # https://github.com/ClangBuiltLinux/linux/issues/1143
```
nickdesaulniers, line 958: do these comments need to be deleted as well?

generator.yml
```diff
@@ -1002,7 +996,7 @@ builds:
   - {<< : *ppc64le,           << : *stable-5_10,      << : *llvm,            boot: true,  llvm_version: *llvm_11}
   # https://github.com/ClangBuiltLinux/linux/issues/1023
   # https://github.com/ClangBuiltLinux/linux/issues/1143
```
nickdesaulniers, line 998: same

nickdesaulniers (approved):
thanks for the patches! 🍏 🌴

generator.yml
nathanchance: Yes, good call!

generator.yml
nathanchance: This one is fine because I did not touch 5.10.

nathanchance:
Thanks for the review!

PR #232, opened 2021-11-04
Apply a patch to unbreak the Thumb2 build

nathanchance:
See the commit message in the patch for the full description.


nickdesaulniers (approved):
thanks for the fix 🥼 🚯

nathanchance:
Thanks for the review!

PR #233, opened 2021-11-05
Update patches

nathanchance:
Each commit should be self explanatory. Hopefully we will be able to drop these relatively quickly.


nickdesaulniers (approved):
thanks for the fixes! 🇽🇰 🔪

PR #236, opened 2021-11-08
Update patches

nathanchance:
Each commit should be self explanatory, let me know if there are any issues!


nickdesaulniers (approved):
looks great! thanks! :name_badge: :woozy_face:

nathanchance:
Thank you for the review!

Issue #235, opened 2021-11-05
extra newlines printed when printing clang info


nickdesaulniers:
example: https://github.com/ClangBuiltLinux/continuous-integration2/runs/4112552575?check_suite_focus=true
```
Printing clang-nightly checkout date and hash

current date: Fri Nov  5 01:54:05 UTC 2021

clang checkout date: Tue Nov  2 07:17:12 UTC 2021

clang revision: 869989dd713d

clang revision link: https://github.com/llvm/llvm-project/commit/869989dd713d

5 warnings, 1 errors
```

nathanchance:
Are you referring to the extra space between `Nov` and `5 01...`? That is just how `date` is but I guess I can just `sed` it out.

nickdesaulniers:
@nathanchance sorry, not whitespace, I meant newlines.

nathanchance:
Ah. I guess I find it easier to read with spaces between each item but I can remove that if you don't.

PR #237, opened 2021-11-08
parse-debian-clang.sh: Remove newlines between information print outs

nathanchance:
Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/235


nickdesaulniers (approved):
thanks! :o: :coat:

PR #238, opened 2021-11-09
patches: next: Remove return type error patch

nathanchance:
A better patch from Arnd was merged, which causes my patch to break the
build:

https://lore.kernel.org/r/20211105075119.2327190-1-arnd@kernel.org/


nathanchance:
I have pushed another fix because our latest -next run failed to start due to a patch we were carrying being merged.

nickdesaulniers (approved):
keeping -next green is _indeed_ sisyphean! Thanks for the updates! 🧗 🪨

PR #239, opened 2021-11-11
patches: Remove android-mainline

nathanchance:
The one patch we were carrying has been merged as part of the mainline merge.


nickdesaulniers (approved):
thanks! 🏞️ 🥼

PR #243, opened 2021-11-15
Fix 5.15 builds

nathanchance:
This should bring 5.15 back to green.


nickdesaulniers (approved):
thanks for the patch! 💯 🩴

PR #244, opened 2021-11-18
Make it green (11-18-2021)

nathanchance:
This series should make all of CI green again. It sucks to continue to take on patches but these are all in flight so we should be able to drop them sooner rather than later (except maybe the `android12-5.10` patches since it seems LTS merges do not happen as consistently as `android13-5.10`, presumably because it is a released branch).


nickdesaulniers (approved):
wow, thanks for all the fixes. 🧪 ⛵ 

https://github.com/ClangBuiltLinux/linux/issues/1502 (just creating the cross reference)

nathanchance:
Thanks for the review!

PR #248, opened 2021-11-22
Remove CROSS_COMPILE_COMPAT and add CONFIG_COMPAT_VDSO=n build

nathanchance:
TuxMake added CROSS_COMPILE_COMPAT to all arm64 make invocations,
meaning that CONFIG_COMPAT_VDSO is now enabled for all arm64 builds by
default. We had previously been adding CROSS_COMPILE_COMPAT to builds we
knew worked. Now, we have to disable CONFIG_COMPAT_VDSO for ones that we
know do not work, which is just 5.4.

This ends up simplifying our matrix as we will automatically get the
compat vDSO for any future builds we enable.

Link: https://gitlab.com/Linaro/tuxmake/-/commit/91cd7f1a85f76b7b120268362664c501157bb90e

### DIFF
--- a/.github/workflows/5.4.yml
+++ b/.github/workflows/5.4.yml
@@ -70,16 +70,16 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _8b6514fc2e63bf7f97f781e252c04550:
+  _b47eaab7c36202f985db8ea18a308034:
     runs-on: ubuntu-20.04
     needs: kick_tuxsuite_defconfigs
-    name: ARCH=arm64 LLVM=1 LLVM_IAS=1 LLVM_VERSION=14 defconfig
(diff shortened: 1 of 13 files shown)


nathanchance:
Small ping @nickdesaulniers (resolves https://github.com/ClangBuiltLinux/continuous-integration2/runs/4361235842?check_suite_focus=true)

nickdesaulniers (approved):
awesome thank you! 🪨 🧻 ✂️ 

also noting that with
commit 3e6f8d1fa18457d54b20917bd9174d27daf09ab9 ("arm64: vdso32: require CROSS_COMPILE_COMPAT for gcc+bfd")

setting CROSS_COMPILE_COMPAT is no longer necessary when using clang+lld. It doesn't hurt though.

nathanchance:
Thank you for all the review as always! 🎉

PR #252, opened 2021-11-30
patches: Drop patch limiting __builtin_thread_pointer() to clang-14

nathanchance:
The backport is available now:

```
$ podman run --rm -ti docker.io/tuxmake/arm_clang-13 clang --version | head -1
Debian clang version 13.0.1-++20211110062941+9dc7d6d5e326-1~exp1~20211110183517.26

$ git log --oneline ed38280d006c^..9dc7d6d5e326
9dc7d6d5e326 [SystemZ] Give the EXRL_Pseudo a size value of 6 bytes.
32bb956916e3 Bad SLPVectorization shufflevector replacement, resulting in write to wrong memory location
93edfb23b18b [SLP][NFC]Add a test to show an issue with incorrectly extracted pointers.
162f3f18c945 [Aarch64] Correct register class for pseudo instructions
8be24d19fefa [MergeICmps] Don't merge icmps derived from pointers with addressspaces
9e084f4194e6 Fix building with GCC 12:
d5159b99105d MLIR can't support -Bsymbolic link option, fail at CMake time with a helpful message instead of broken runtime
ff2cb6e400c3 [clang] Partially revert d8cd7806310c51af912a647a6ca46de62ff13214.
ed38280d006c [ARM] Use hardware TLS register in Thumb2 mode when -mtp=cp15 is passed
```


nickdesaulniers (approved):
LGTM ❌ 🦉

PR #253, opened 2021-12-02
patches: Drop dwc2 floating literal patch

nathanchance:
It has been merged into all of these trees.


nickdesaulniers (approved):
thanks! 🚈 ⚡

PR #254, opened 2021-12-07
Make failures more obvious

nathanchance:
This pull requests makes two changes to try and help triage failures a little bit easier:

1. Attempt to fetch artifacts if there was a 404 error.
2. Fetch the build's `status.json` if the build's `result` value from `tuxsuite`'s `builds.json` is neither `pass` nor `fail` (maybe a race between the `tuxsuite` client/backend?) to try and get an updated result.

Each commit has further information/motivation. This has not seen any formal testing yet but there is no obvious errors that I can see.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -17,7 +17,7 @@ def _fetch(title, url, dest):
     # TODO: use something more robust like python wget library.
     retries = 0
     max_retries = 7
-    retry_codes = [500, 504]
+    retry_codes = [404, 500, 504]
     while retries < max_retries:
         try:
             if retries:
@@ -34,11 +34,6 @@ def _fetch(title, url, dest):
                 print_yellow("%s download error (%d), retrying..." %
                              (title, err.code))
                 pass
-            elif err.code == 404:
-                print_red(
-                    "%s could not be found (404 error), did the build timeout?"
-                    % (title))
-                sys.exit(1)
             else:
                 print_red("%d error trying to download %s" % (err.code, title))
                 sys.exit(1)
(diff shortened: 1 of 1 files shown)


check_logs.py
```diff
@@ -54,6 +54,34 @@ def _fetch(title, url, dest):
...
+            break
+
+        if retries:
+            time.sleep(2**retries)
+        retries += 1
+
+        status_json = "status.json"
+        url = build["download_url"] + status_json
+        _fetch("status.json", url, status_json)
+        build = json.load(open(status_json))
```
nickdesaulniers, line 76: should we be reassigning `build` here? Wont `verify_build` return something different than if `get_build` had simply succeeded the first time? For example, say the initial `get_build` fails, we retry once, and get success. `verify_build` will return status.json rather than one of the objects in build.json. Aren't those different schema?

nathanchance: As far as I understand it, the `builds.json` that we have `tuxsuite` generate is just the collection of the individual `status.json` files from each build. Overriding the `build` variable is intentional in that case.

nathanchance: So it looks like there might be a few more fields that we get from the `tuxsuite` client (comparing https://github.com/ClangBuiltLinux/continuous-integration2/runs/4453112113?check_suite_focus=true with https://builds.tuxbuild.com/21yu3ofQYT1H3nC9EXGtDpMKQ7y/status.json), like `environment`, `finished_time`, `plan`, `user`, and `user_agent`. However, does that matter? We do not use any of those fields.

nickdesaulniers: > the builds.json that we have tuxsuite generate is just the collection of the individual status.json files from each build

oh, it is! I never noticed that!

nickdesaulniers (approved):
LGTM thanks! 🤖 🪨

nathanchance:
Thanks for the review!

PR #255, opened 2021-12-08
Drop -next clang-10 builds

nathanchance:
The minimum supported version of LLVM has been raised to 11.0.0 so these
builds will no longer work.


nickdesaulniers (approved):
LGTM 🧑‍⚕️ 🛷

PR #256, opened 2021-12-09
Change TuxSuite targets

nathanchance:
This series updates the targets that we call for TuxSuite, which will save us a little bit of work (maybe more for modules). See the individual commits for each change to understand the rationale.


nickdesaulniers (approved):
nice series! this should save quite a bit of build minutes! 💵 💲

nathanchance:
Thanks for the review!

PR #257, opened 2021-12-09
Improve check_logs.py output

nathanchance:
A couple of small improvements around `check_logs.py` that should help with triaging issues that arise.


nickdesaulniers (approved):
LGTM :lab_coat: :waffle:

PR #258, opened 2021-12-10
Update patches (12-10-2021)

nathanchance:
See individual commits for details.


nickdesaulniers (commented):
do you need to run regen when the series files are removed or created?

nathanchance:
> do you need to run regen when the series files are removed or created?

No, only when fully dropping series; this PR does not do that. I added a check in `check-patches.sh` to account for this. `./generate.sh all` does not change anything.

nathanchance:
Thank you for the review as always!

Issue #261, opened 2021-12-20
fail better when patches fail to apply


nickdesaulniers:
See [discussion](https://github.com/ClangBuiltLinux/continuous-integration2/pull/260#issuecomment-998229537).

PR #262, opened 2021-12-20
check_logs.py: Print build log and exit when patch fails to apply

nathanchance:
TuxSuite's build.log shows the output of patch when a patch fails to
apply; print that log so that we can easily triage patch failures.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/261

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -82,6 +82,13 @@ def verify_build():
         print_red("status.json did not give a pass/fail result!")
         sys.exit(1)
 
+    if build["status_message"] == "Unable to apply kernel patch":
+        print_red(
+            "Patch failed to apply to current kernel tree, does it need to be removed or updated?"
+        )
+        fetch_logs(build)
+        sys.exit(1)
+
     return build
 
 


check_logs.py
```diff
@@ -258,6 +267,7 @@ def boot_test(build):
         show_builds()
         sys.exit(1)
     build = verify_build()
+    check_patch_failure(build)
```
nickdesaulniers, line 270: We already check `build["result"] == "fail"` in `verify_build`; let's move this check into `verify_build`. Otherwise this will still retry. If we already know that the the status message is due to patch application failure, then we should not retry.

nathanchance: I am not sure I follow the "still retry" logic but sure, done. PTAL.

check_logs.py
```diff
@@ -82,6 +82,13 @@ def verify_build():
         print_red("status.json did not give a pass/fail result!")
         sys.exit(1)
 
+    if build["status_message"] == "Unable to apply kernel patch":
```
nickdesaulniers, line 85: Looking at [an example](https://github.com/ClangBuiltLinux/continuous-integration2/runs/4566921218?check_suite_focus=true), it looks like when `"status_message"` == "Unable to apply kernel patch"` that `"result" == "fail"`.

Let's say the first request we make has `"status_message"` == "Unable to apply kernel patch"` and `"result" == "fail"`. In such a case, we will retry `max_retries` times, then exit (L83).

I think we need to check this either before the retry loop, or within the retry loop.

nathanchance: > Looking at [an example](https://github.com/ClangBuiltLinux/continuous-integration2/runs/4566921218?check_suite_focus=true), it looks like when `"status_message" == "Unable to apply kernel patch"` AND `"result" == "fail"`.

Correct, that should always be what it is.

> Let's say the first request we make has `"status_message" == "Unable to apply kernel patch"` and `"result" == "fail"`. In such a case, we will retry `max_retries` times, then exit (L83).

Shouldn't we bail out of the while loop before getting to `max_retries` because `"result" == "fail"`?

```
    while retries < max_retries:
        if build["result"] == "fail" or build["result"] == "pass":
            break
```

I do not necessarily mind moving it earlier but I don't think there will ever be a time where `"status_message"` is `"Unable to apply kernel patch"` and `"result"` is not `"fail"`.

nickdesaulniers: > Shouldn't we bail out of the while loop before getting to max_retries because "result" == "fail"?

Yes, but then we'll exit (L83) BEFORE checking "status_message" in that case. The code you're adding in this PR looks like it will be unreachable to me.

nickdesaulniers: ah, no it's not nvm.

nickdesaulniers (approved):

nathanchance:
Thanks for double checking me!

PR #266, opened 2021-12-31
Move to a container based boot workflow

nathanchance:
With GitHub Actions, we can run jobs in a container, meaning that we do not have to be tied to Ubuntu's QEMU versions. This allows us to use a more recent operating system, such as a rolling release option like Arch Linux, with little effort.

See each individual commit for the logic.

The container repository is at https://github.com/ClangBuiltLinux/containers.

This has not been tested, as I am not sure of the best way to do that.


nickdesaulniers (approved):
LGTM; thanks for the patch!

just be ready to revert/clean up any fallout after merging this. :boat: :ice_cube:

nathanchance:
Sure thing! Mainline has been consistently green for the past couple of days so if it goes red, I'll immediate triage and revert as necessary. Thanks for the review.

PR #270, opened 2022-01-06
patches: android12-5.10: Add patch to fix Thumb2 kernel boot on newer QEMU

nathanchance:
This is in 5.10.77 but that version has not been merged into
android12-5.10 yet (as it is a released branch), which is currently at
5.10.66.

Apply this patch in the meantime so that the kernel continues to boot.


nickdesaulniers (approved):
thanks for finding this! LGTM 💯 👌

PR #273, opened 2022-01-11
patches: Drop 5.15 and 5.10

nathanchance:
This patch has been merged into 5.15.14 and 5.10.91:

https://lore.kernel.org/r/1641912099199218@kroah.com/
https://lore.kernel.org/r/1641912091110251@kroah.com/


nickdesaulniers (approved):
thanks for the PR! 🅾️ 🐂

PR #274, opened 2022-01-11
patches: Apply hack patch to fix the build on android-mainline

nathanchance:
I have submitted a pull request to TuxMake to upgrade its AOSP LLVM
version but until that is submitted, we need this hack to fix the build.
It should only impact AOSP LLVM, rather than the other clang versions.

Link: https://github.com/ClangBuiltLinux/linux/issues/1561
Link: https://gitlab.com/Linaro/tuxmake/-/merge_requests/227


nickdesaulniers (approved):
I like it

consider waiting to merge this in case @vishalbhoj and team update the AOSP LLVM version today 🤞

nathanchance:
> consider waiting to merge this in case @vishalbhoj and team update the AOSP LLVM version today 🤞

Sure. It might take a bit for the containers to update even after my patch is merged, between the push to Docker and the mirroring to TuxSuite's backend so I think we will still need this for at least a couple of days but I'll wait a couple of hours before merging it in case it gets fast tracked.

PR #276, opened 2022-01-12
patches: Add a series to fix ARCH=arm64 with mainline and next

nathanchance:
This is not a clang specific build failure but it does impact our
ARCH=arm64 allmodconfig builds.

Link: https://lore.kernel.org/r/Yd2jcfXqqKHK%2FNaG@archlinux-ax161/
Link: https://lore.kernel.org/r/20220111172919.6567-1-zajec5@gmail.com/


nickdesaulniers (approved):
thanks for keeping these green. surprising it got through to mainline 🔥 🗑️

PR #277, opened 2022-01-13
check-patches.sh: Check for patches but no '--patch-series' 

nathanchance:
PR #276 added some patches but forgot to regenerate the workflow files
so they were not applied.

Add a check for this so that presubmit testing catches it.

Regenerate the workflow files so that the patches get applied properly.


nickdesaulniers (approved):
ah I missed that, too. This is a good idea!

I think removing+adding `--patch-series` causes a fair amount of churn. I'm curious if `tuxsuite` will error if there's no patch files in that dir? If not, we could probably just leave the flag specified always.

nathanchance:
> I think removing+adding `--patch-series` causes a fair amount of churn. I'm curious if `tuxsuite` will error if there's no patch files in that dir? If not, we could probably just leave the flag specified always.

I am not sure, it probably depends on how `tuxsuite` applies the patches. We could potentially just keep an empty `series` file in the folder when there are no patches but regenerating the files is cheap so at the moment, I am inclined to just leave this as is.

PR #280, opened 2022-01-14
patches/mainline: Add a SCSI patch from -next

nathanchance:
This resolves a -Wuninitialized warning, which breaks our hexagon
allmodconfig build because -Werror is enabled (to catch problems like
this).

https://builds.tuxbuild.com/23hS2caXpXboLsIdKN7hdaLThbt/build.log

This patch is in the 5.17 branch so it should make it to Linus soon.


nathanchance:
I am going to merge this without approval to keep the builds green over the weekend. If anyone disagrees, scream and I can revert or talk other solutions.

PR #285, opened 2022-01-25
Disable CONFIG_DEBUG_INFO_BTF for gki_defconfig

nathanchance:
There are issues with the pahole version that TuxMake ships, which is
causing infrastructure issues. Until a fix can be developed, disable
BTF generation so that pahole does not run.


nickdesaulniers (approved):
thanks for a quick fix! 🍍 🌭

PR #288, opened 2022-01-26
Add pinctrl-thunderbay.c patches to arm64-fixes

nathanchance:
This mainline build breakage has been inherited by the arm64-fixes tree,
as it is based on v5.17-rc1.

Add the patches to keep it building and potentially catch other
breakage.


nickdesaulniers (approved):
thank you! 🐗 🛥️

PR #289, opened 2022-01-26
Add blake2s CFI workaround patch for android-mainline

nathanchance:
This is needed to avoid the panic on boot mentioned in the commit
message of the patch, which occurs with the arm64 gki_defconfig.


nickdesaulniers (approved):
LGTM 🥥 😩

Issue #290, opened 2022-01-27
INT DW_ATE_unsigned_56 Error emitting BTF type


nathanchance:
Our `clang-nightly` builds using Fedora's configurations recently started failing during BTF generation with:

```
[635133] INT DW_ATE_unsigned_56 Error emitting BTF type
Encountered error while encoding BTF.
FAILED: load BTF from vmlinux: Invalid argument
```

arm64: https://builds.tuxbuild.com/24HTWG82iJRAs7sVbMxgydiKuVY/build.log
powerpc: https://builds.tuxbuild.com/24HTWLTLyfUnM6rVmcFfkIIwi06/build.log
s390x: https://builds.tuxbuild.com/24HTWKDJkNLyF13FYx0NBdvn9LM/build.log
x86_64: https://builds.tuxbuild.com/24HTWMan33SRUw1NLGs6RbOOcg2/build.log

This happens because the default DWARF version for `clang-nightly` has [been changed to DWARF5](https://github.com/llvm/llvm-project/commit/d3b26dea16108c427b19b5480c9edc76edf8f5b4) (Fedora's configs have `CONFIG_DEBUG_INFO_DWARF_TOOLCHAIN_DEFAULT=y`) and `pahole` had [a bug fix for `clang` + DWARF5](https://git.kernel.org/pub/scm/devel/pahole/pahole.git/commit/?id=7d8e829f636f47aba2e1b6eda57e74d8e31f733c) in `1.21` but TuxMake uses `1.19`.

There is some talk of bumping the version of `pahole` in the TuxMake containers to fix another breakage in Android, which would resolve this.

Alternatively, the Fedora kernel team could potentially opt in to DWARF4 explicitly (or we could do it in CI) but DWARF5 has several benefits over DWARF4 so it might be nice to keep it.

We can probably remove the `depends on !DEBUG_INFO_BTF` from `config DEBUG_INFO_DWARF5` in `lib/Kconfig.debug` if `pahole` is `1.21` or newer, but that would not solve this issue, as the switch from DWARF4 to DWARF5 was in the toolchain, not Kconfig.

PR #291, opened 2022-01-28
Disable CONFIG_DEBUG_INFO_BTF for Fedora configs

nathanchance:
Until this can get sorted out on the TuxSuite side.

Link: https://github.com/ClangBuiltLinux/continuous-integration2/issues/290

### DIFF
--- a/.github/workflows/5.15.yml
+++ b/.github/workflows/5.15.yml
@@ -2233,15 +2233,15 @@ jobs:
         path: builds.json
         name: output_artifact_distribution_configs
         if-no-files-found: error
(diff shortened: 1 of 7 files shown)


nickdesaulniers (approved):
Is this needed for allmodconfig builds too, or just Fedora configs?

nathanchance:
> Is this needed for allmodconfig builds too, or just Fedora configs?

`CONFIG_DEBUG_INFO` does not get enabled with `allmodconfig` without explicit opt-in. As far as I am aware, the Fedora configurations are the only ones that show this right now but I'll keep an eye out in case we have to keep playing whack-a-mole.

PR #292, opened 2022-01-29
Fix CONFIG_DEBUG_INFO_BTF=n for Fedora configs

nathanchance:
Fedora's 32-bit ARM config does not have CONFIG_DEBUG_INFO_BTF enabled
so it does not need CONFIG_DEBUG_INFO_BTF=n.

I missed the PowerPC configuration by mistake.


nathanchance:
I am going to merge this without approval so that the build stays green over the weekend.

PR #294, opened 2022-02-02
patches: Remove mainline series

nathanchance:
These two patches are now merged into mainline:

https://git.kernel.org/linus/3e5832e923a9c3a12c76980f68853668d4675ecf


nickdesaulniers (approved):
I'm happy to be back to 0 patches to mainline. 🪢 🪵

nathanchance:
Ack, I will split that out in the future. I just lumped the result of `generate.sh all` in this commit but having the LLVM_TOT_VERSION bump seperately would definitely be helpful for review. Thanks for the review!

nathanchance:
Ha, that was supposed to be on #296 obviously... got my tabs switched up :)

Issue #295, opened 2022-02-02
Add support for LLVM 14


nathanchance:
The `release/14.x` branch has been cut meaning `clang-nightly` will now be `clang-15` and there will eventually be `clang-14` through [apt.llvm.org](https://apt.llvm.org/). The transition plan:

- [ ] Wait for LLVM 14 binaries on [apt.llvm.org](https://apt.llvm.org/)
- [ ] Report availability of LLVM 14 binaries on [apt.llvm.org](https://apt.llvm.org/) to [TuxMake issue tracker](https://gitlab.com/Linaro/tuxmake)
- [ ] Wait for TuxMake to provide `clang-14` images
- [ ] Wait for TuxSuite to allow `clang-14` to be specified
- [ ] Update `generator.yml` to provide `llvm_13` anchor and copy all `llvm_latest` builds to it
- [ ] Update `generator.yml` to shift `llvm_latest` to 14 and copy all `llvm_tot` builds to it

NOTE: We will likely hit GitHub's matrix limit on at least -next and mainline with this shift so we will need to think about how to handle that.

sylvestre:
Working on the apt.llvm.org packages now! Should be ready in the next few days :)

sylvestre:
> Wait for LLVM 14 binaries on [apt.llvm.org](https://apt.llvm.org/)

done!

nathanchance:
Thanks a lot! I have filed https://gitlab.com/Linaro/tuxmake/-/issues/167 so that they can be added to `tuxmake` / `tuxsuite`.

nickdesaulniers:
Since we'll hit the gh action job limit, we should first change how we generate workflow files; perhaps a {kernel_tree}-{llvm version} workflow. That should be ~95 workflow files (if I did the math right).  Probably can generate the markdown for tiles, too.

nathanchance:
Support is now available in `tuxsuite`: https://gitlab.com/Linaro/tuxsuite/-/merge_requests/215

I should be able to work on splitting up the TuxSuite and workflow files today or tomorrow.

nickdesaulniers:
nice job working through all the motions

PR #296, opened 2022-02-02
Bump llvm_tot anchor and LLVM_TOT_VERSION

nathanchance:
This bumps our llvm_tot anchor and the LLVM_TOT_VERSION file to 15, as the version has been bumped upstream: https://github.com/llvm/llvm-project/commit/a2601c98873376bbbeff4b6eddf0f4d920535f8b

This does not add LLVM 14 builds yet, as we need apt.llvm.org to add LLVM 14 packages and TuxMake to support clang-14 as a target. I have filed an issue to track this: https://github.com/ClangBuiltLinux/continuous-integration2/issues/295


nickdesaulniers (approved):
The change to `LLVM_TOT_VERSION` was harder to notice in the final commit; I wouldn't mind if that was broken out into a distinct commit next time. LGTM though, so thank you. 🧥 🇨🇷

PR #299, opened 2022-02-09
Enable CONFIG_DEBUG_INFO_BTF for Android and Fedora configs

nathanchance:
pahole 1.22 has been backported to the tuxmake images so there should be
no more build failures.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/290

### DIFF
--- a/.github/workflows/5.15.yml
+++ b/.github/workflows/5.15.yml
@@ -2271,15 +2271,15 @@ jobs:
         name: output_artifact_distribution_configs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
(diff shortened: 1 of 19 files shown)


nathanchance:
For what it's worth, it seems like Android is going to turn the configuration off again: https://android-review.googlesource.com/c/kernel/common/+/1977410

A better title for this would probably have been "Stop disabling CONFIG_DEBUG_INFO_BTF" but oh well :)

nickdesaulniers (approved):
LGTM thanks!
🍸 🌅

PR #300, opened 2022-02-14
Run generate.sh on pull requests and pushes

nathanchance:
The first change is just a small clean up to make these checking workflows a little bit more maintainable.

The second change runs `generate.sh` on every push and pull request so that we can make sure that our GitHub Actions and TuxSuite files are always updated.


nickdesaulniers (approved):
LGTM thanks! 🏖️ 🛷

PR #305, opened 2022-02-17
Include toolchain version in workflow name

nathanchance:
Otherwise, the workflow names are all the same in GitHub's UI, which
makes it hard to navigate.


nickdesaulniers (approved):
nice touch! 🛷 😪

PR #306, opened 2022-02-17
Fix CI badges

nathanchance:
Per GitHub's own documentation, the syntax for the badge image is

```
https://github.com/<OWNER>/<REPOSITORY>/actions/workflows/<WORKFLOW_FILE>/badge.svg
```

Currently, ours is

```
https://github.com/<OWNER>/<REPOSITORY>/workflows/<WORKFLOW_FILE_MINUS_EXTENSION>/badge.svg
```

To follow GitHub's own documentation, use the filename plus extension.

After the split of the workflow files and the renaming in #305, the
links will no longer show just the workflow in question. Fix this by
grabbing the name of the workflow from the "name:" at the top of the
file and make it URL friendly with + instead of " ".

Link: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

NOTE: This is based on #305, due to the renaming of the workflows.


nathanchance:
Rebased on main after #305 was merged and it turns out the URL can get a lot simpler so I did that. I am going to merge it now, holler if you don't like the change. Thanks for the review as always!

PR #307, opened 2022-02-17
Move shell scripts to scripts/

nathanchance:
There are a lot of files in the root of the directory. To help organize
things a bit, move all shell scripts into a scripts/ folder and adjust
the paths accordingly.

NOTE: This change is fairly opinionated, I am fine with discarding it if you
feel all the script should remain in the root folder.


nickdesaulniers (approved):
LGTM thanks! :oncoming_bus: :fire:

PR #309, opened 2022-02-17
Move and update cron logic

nathanchance:
This PR moves the cron schedule logic into Python and updates it to do builds less frequently and at different intervals.


nickdesaulniers:
almost needs a double keyed dictionary

['tree']['llvm_version'] = schedule

I wonder if there's a way to encode that in yaml that can simply be looked up in python given the tree and llvm version?

nathanchance:
Closing in favor of #310.

PR #310, opened 2022-02-18
Update cron schedule logic

nathanchance:
This is an alternative to #309.

Instead of moving the cron schedule logic into `generate_workflow.py`, it stays in `generator.yml`, which is more verbose and declarative, but gives us more accurate control over the schedule.

### DIFF
--- a/generator.yml
+++ b/generator.yml
@@ -28,10 +28,10 @@ urls:
   - &x86_64-fedora-config-url  https://src.fedoraproject.org/rpms/kernel/raw/rawhide/f/kernel-x86_64-fedora.config
   - &x86_64-suse-config-url    https://github.com/openSUSE/kernel-source/raw/master/config/x86_64/default
 schedules:
-  - &six_hours {schedule: "0 0,6,12,18 * * *"}
-  - &daily     {schedule: "0 0 * * *"}
+  - &six_hours      {schedule: "0 0,6,12,18 * * *"}
+  - &daily_midnight {schedule: "0 0 * * *"}
   # -next updates M-F in the evening AEST, which is usually around 12:00PM UTC
-  - &weekdays  {schedule: "0 12 * * 1,2,3,4,5"}
+  - &weekdays       {schedule: "0 12 * * 1,2,3,4,5"}
 trees:
   - &mainline         {git_repo: *mainline-url, git_ref: master,              name: mainline}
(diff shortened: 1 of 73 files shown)


generator.yml
```diff
@@ -33,25 +33,114 @@ schedules:
...
+  - &android12-5_10   {git_repo: *android-url,  git_ref: android12-5.10,      name: android12-5.10}
+  - &android12-5_4    {git_repo: *android-url,  git_ref: android12-5.4,       name: android12-5.4}
+  - &android-4_19     {git_repo: *android-url,  git_ref: android-4.19-stable, name: android-4.19}
+  - &android-4_14     {git_repo: *android-url,  git_ref: android-4.14-stable, name: android-4.14}
+  - &android-4_9      {git_repo: *android-url,  git_ref: android-4.9-q,       name: android-4.9}
+  - &tip              {git_repo: *tip-url,      git_ref: master,              name: tip}
+  - &arm64-core       {git_repo: *arm64-url,    git_ref: for-next/core,       name: arm64}
+  - &arm64-fixes      {git_repo: *arm64-url,    git_ref: for-next/fixes,      name: arm64-fixes}
+tree_schedules:
+  - &mainline_llvm_tot             {llvm_version: *llvm_tot,     << : *mainline,         << : *six_hours}
```
nickdesaulniers, line 61: can we add the key `llvm_version` to the `llvm_versions` entries, that way in this `tree_schedules` table we could have entries like:

```yaml
  - &mainline_llvm_tot { << : *llvm_tot, << : *mainline, << : *six_hours}
```
so that `llvm_versions` is a list of objects rather than a list of integers?

nickdesaulniers (approved):
I much prefer this approach to #309 . Thank you for taking the time to try both approaches.

I think we might be able to factor the table further, making it even more concise and readable, but I don't care whether you do it in this patch/series, or after.

Thanks for the work on this! :flight_departure: :flight_arrival:

generator.yml
nathanchance: Yes, I think that should be possible... Let me try it and see what breaks :)

nathanchance: Done, PTAL!

nickdesaulniers (approved):
nice work! I appreciate how readable (IMO) the table of schedules is. :owl: :cl:

nathanchance:
Thanks for the review!

PR #313, opened 2022-02-23
patches: next: Add a patch to fix build with KVM_WERROR

nathanchance:
Our allmodconfig ThinLTO builds disables `KASAN`, which causes `KVM_WERROR`
to be enabled due to the `default y if !KASAN`.


nathanchance:
Checks passed, merging without explicit approval to keep the build green.

PR #317, opened 2022-03-01
Fix up check_logs.py and messages/documentation after scripts move

nathanchance:
This fixes a few places where `.sh` scripts are being referenced by the old location, namely `check_logs.py`, which is causing the CI to fail currently.

NOTE: I intend to merge this without explicit approval once the checks pass to attempt to prevent the CI from going completely red.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,10 +109,10 @@ to it. Ex.
 
 ```
 # Generate just next and mainline TuxSuite and GitHub Action workflows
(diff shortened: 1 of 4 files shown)


nathanchance:
It appears these checks won't run for some time due to all the running TuxSuite workflows. I have run them locally and they pass so I am merging. Please yell if there is something wrong and I will fix it in the morning.

PR #318, opened 2022-03-03
patches: next: Replace warning hiding patch with proper fix

nathanchance:
The hack patch fails to apply as of next-20220302. Replace it with a
solution that can be accepted upstream.


nathanchance:
Checks all passed. I am merging this without approval to try and keep the CI green for `next-20220303`. Please yell if there are objections after the fact!

PR #320, opened 2022-03-03
Drop android-4.14 hack patch for ld.lld

nathanchance:
This is no longer necessary, as ld.lld has been fixed in both main and
release/14.x.


nickdesaulniers (approved):
LGTM thanks! 🥼 🍖

PR #325, opened 2022-03-07
patches: next: Drop KVM patch

nathanchance:
It is available in next-20220307.


nickdesaulniers (approved):
LGTM thanks :ok: :llama:

Issue #265, opened 2021-12-28
test PPC with LLVM_IAS=1


nickdesaulniers:
@aik [sent a series](https://lore.kernel.org/llvm/20211221055904.555763-1-aik@ozlabs.ru/) for enabling LLVM_IAS=1. @mpe [accepted it](https://git.kernel.org/pub/scm/linux/kernel/git/powerpc/linux.git/log/?h=next), so it should be in linux-next soon.  The cover letter mentions we need https://reviews.llvm.org/D115419 that @nemanjai is working on. As soon as that lands, we should be able to start testing probably our ppc64le configs on linux-next with clang-14 with LLVM_IAS=1 enabled.

mpe:
It also needs [the hwsync suport](https://reviews.llvm.org/rG2fb9029f2678e3698de202fa88150b3b03cf76da) which is already merged in LLVM.

nathanchance:
With [D115419](https://reviews.llvm.org/D115419) applied and Linux at commit [112450df61b7](https://git.kernel.org/linus/112450df61b7373529b0fe4c122ad13b89d80a8a) ("Merge branch 'i2c/for-mergewindow' of git://git.kernel.org/pub/scm/linux/kernel/git/wsa/linux"), we get closer to building a `powernv_defconfig` kernel with `LLVM=1`:

```
$ make -skj"$(nproc)" ARCH=powerpc LLVM=1 powernv_defconfig all
/usr/sbin/ld: unrecognised emulation mode: elf64lppc
Supported emulations: elf_x86_64 elf32_x86_64 elf_i386 elf_iamcu elf_l1om elf_k1om
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [arch/powerpc/kernel/vdso64/Makefile:44: arch/powerpc/kernel/vdso64/vdso64.so.dbg] Error 1
...
```

This is https://github.com/ClangBuiltLinux/linux/issues/774. If I quickly hack around that error:

```diff
diff --git a/arch/powerpc/kernel/vdso64/Makefile b/arch/powerpc/kernel/vdso64/Makefile
index 3c5baaa6f1e7..45dc62ba1541 100644
--- a/arch/powerpc/kernel/vdso64/Makefile
+++ b/arch/powerpc/kernel/vdso64/Makefile
@@ -39,6 +39,8 @@ asflags-y := -D__VDSO64__ -s
 targets += vdso64.lds
[...]

PR #326, opened 2022-03-08
[v2] Build most powerpc64le configs with LLVM=1 LLVM_IAS=1

nathanchance:
On mainline, SUSE's configuration needs to build with LLVM_IAS=0 because
of https://github.com/ClangBuiltLinux/linux/issues/1418.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/265


nathanchance:
Now that we have a solution to https://github.com/ClangBuiltLinux/linux/issues/1581, I have applied it so that we can use the first version of the pull, which won't have the boot failures.

nathanchance:
Alright, `next-20220310` boots without any patches so I have applied Alexey's patch to mainline. We can drop it once it is merged.

nathanchance:
I added the patch that drops the `stabs` annotations so that all of these configurations can be built with `LLVM=1 LLVM_IAS=1`. We should only have to carry that patch for a couple of weeks. This should be ready to go in.

nickdesaulniers (approved):
LGTM; thanks! :palm_tree: :cook_islands:

PR #329, opened 2022-03-11
check_logs.py: Use proper syntax for removing problem matcher

nathanchance:
The oopses problem matcher is not getting removed for KUnit runs as it
should:

https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/1964728851

According to GitHub's official documentation, there should be two colons
at the end of the remove-matcher command but they are not present in the
check_logs.py print out. Add them to hopefully fix the problem matcher
getting disabled.

Link: https://github.com/actions/toolkit/blob/main/docs/commands.md#problem-matchers


nickdesaulniers (approved):
LGTM; thanks! :lab_coat: :llama:

Issue #327, opened 2022-03-10
problem with matchers


nickdesaulniers:
looking at https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/1964728851, if you scroll way down, you'll see warnings like:
```
TCG doesn't support requested feature: CPUID.80000001H:EDX.syscall [bit 11]
```
and
```
Call Trace
```
not sure where these are coming from, exactly, but they don't look like lines in the build logs?

nathanchance:
Those both come from QEMU. The problem matchers match against the whole `check_logs.py` step, which includes both the build logs and QEMU.

The `Call Trace` was specifically added in https://github.com/ClangBuiltLinux/continuous-integration2/commit/500f88e0e5dd5f4830bbb5e9c34316e863f956fb.

The TCG warning is just a byproduct of matching on `warning:`, which we might be able to clean up.

nathanchance:
So the `Call Trace` highlighting should not actually happen in those logs. I think that should be fixed with https://github.com/ClangBuiltLinux/continuous-integration2/pull/329.

The TCG warning is likely due to our use of `-cpu Nehalem` but I am not really sure. Perhaps we should use a different architecture for `-a x86` in boot-utils? I'll see if I can play around with this.

PR #330, opened 2022-03-17
boot-utils: Update to latest main branch

nathanchance:
We will need the recently merged `lpa2=off` when QEMU 7.0.0 is released.


nathanchance:
Ah, I forgot that this also resolves https://github.com/ClangBuiltLinux/continuous-integration2/issues/327 fully (first part was https://github.com/ClangBuiltLinux/continuous-integration2/pull/329).

PR #331, opened 2022-03-18
Drop arm64-fixes patches

nathanchance:
The arm64-fixes tree has been fast forwarded to 5.17-rc8, which already
contains these patches.


nickdesaulniers (approved):
LGTM thanks! :ox: :llama:

PR #335, opened 2022-03-21
patches: Drop -next

nathanchance:
The one patch we are carrying has been merged into drm-misc-next-fixes,
which will be included in the next -next version.

Link: https://cgit.freedesktop.org/drm/drm-misc/commit/?id=a860f266a0e19f271b839451d291a6acf6ddcfe8


nickdesaulniers (approved):
LGTM thanks! :llama: :athletic_shoe:

PR #336, opened 2022-03-21
Test the latest stable release

nathanchance:
This series adds support for testing the latest stable release, currently Linux 5.17. Nothing super special here, we are just increasing our coverage to ensure the stable releases stay building, regardless of long term support status.

NOTE: `linux-5.17.y` does not exist in [the stable repository](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/) but I expect it to be added soon.


nickdesaulniers (approved):
LGTM thanks! :lobster: :sa:

PR #338, opened 2022-03-23
Revert 4eec5c448d09646659806afdd42b080a7002620f for stable

nathanchance:
We are applying patches to make 4eec5c448d09646659806afdd42b080a7002620f work in mainline. Given that the patches are unlikely to be stable material when merged, just revert 4eec5c448d09646659806afdd42b080a7002620f for stable.

Link: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/2028368723


nickdesaulniers (approved):
that's ok, thanks for the patches. 🔋 💻

PR #339, opened 2022-03-25
Disable ARCH=um builds for now

nathanchance:
`tuxsuite` needs to support `ARCH=um`.

Link: https://gitlab.com/Linaro/tuxsuite/-/issues/157


nathanchance:

PR #340, opened 2022-03-25
Update AOSP LLVM builds

nathanchance:
We can re-enable the disabled ARM builds because the compiler has been upgraded.

With the upgrade, we now trigger the ld.lld .plt bug for CONFIG_X86_KERNEL_IBT, so the builds that enable that config need to be disabled until the next compiler upgrade.


nathanchance:

PR #342, opened 2022-03-25
patches: mainline: Add -Wimplicit-fallthrough fix for KVM

nathanchance:
KVM can enable -Werror independent of CONFIG_WERROR, which happens with
our allmodconfig + ThinLTO builds. Keep them green by applying this
already queued patch.


nathanchance:
I am going to submitted a different version of this.

PR #343, opened 2022-03-25
Update mainline patches

nathanchance:
KVM can enable -Werror independent of CONFIG_WERROR, which happens with
our allmodconfig + ThinLTO builds. Keep them green by applying an
already queued patch.

The PowerPC patches have been merged:

https://git.kernel.org/linus/1f1c153e406a7375ae0fc3d6000b64e7ba27cf8a


nathanchance:

PR #344, opened 2022-03-26
Disable ARCH=arm allyesconfig on mainline

nathanchance:
Same as a9d8571b91a5b19cd6a48780ed2231938a25283f, but for mainline.


nathanchance:

PR #345, opened 2022-03-29
Remove Fedora i686 configuration

nathanchance:
This configuration has been dropped, as support for 32-bit x86 was
dropped in Fedora 31.

We do not actually lose much coverage, as we had this configuration
disabled for most recent clang versions due to
https://github.com/ClangBuiltLinux/linux/issues/1442.

Link: https://src.fedoraproject.org/rpms/kernel/c/ec96983c95a52f6df6e20e276ff9c761139e32d8
Link: https://fedoraproject.org/wiki/Changes/Stop_Building_i686_Kernels

### DIFF
--- a/.github/workflows/5.15-clang-11.yml
+++ b/.github/workflows/5.15-clang-11.yml
@@ -599,25 +599,6 @@ jobs:
         name: output_artifact_distribution_configs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _09238b29fc27aba9cd6fbf0fc8ae74fa:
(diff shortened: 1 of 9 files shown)


nickdesaulniers (approved):
thanks for noticing Fedora spun these down!
:clamp: :leafy_green:

nathanchance:
It helps that the build errored :)

https://github.com/ClangBuiltLinux/continuous-integration2/runs/5743129779?check_suite_focus=true

Thanks for the review as always!

PR #351, opened 2022-04-06
Update mainline patches (April 6, 2022)

nathanchance:
Add a btrfs patch to fix a compile warning/error with ToT LLVM.

Remove a KVM patch that has already been applied to mainline.


nickdesaulniers (approved):
LGTM thanks! :ox: :llama:

PR #353, opened 2022-04-07
Fix ARCH=um boot testing, part 2

nathanchance:
When using Docker, `/dev/shm` is mounted as `noexec`, which prevents UML
from working properly:

```
Checking if /dev/shm is on tmpfs...OK
Checking PROT_EXEC mmap in /dev/shm...Operation not permitted
/dev/shm must be not mounted noexec

shm on /dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k,inode64)
```

To remove this limitation, the `--ipc` flag can be used with the `host`
value, which allows UML to work on GitHub Actions:

```
Checking that ptrace can change system call numbers...OK
Checking syscall emulation patch for ptrace...OK
Checking advanced syscall emulation patch for ptrace...OK
Checking environment variables for a tempdir...none found
Checking if /dev/shm is on tmpfs...OK
Checking PROT_EXEC mmap in /dev/shm...OK
Adding 18731008 bytes to physical memory to account for exec-shield gap
Linux version 5.18.0-rc1 (tuxmake@tuxmake) (Debian clang version 15.0.0-++20220405074332+72fe439a4e11-1~exp1~20220405074431.211, Debian LLD 15.0.0) #1 @1649311859

tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev,inode64)
```

Link: https://stackoverflow.com/questions/54729130/how-to-mount-docker-tmpfs-with-exec-rw-flags

### DIFF
--- a/.github/workflows/4.14-clang-13.yml
+++ b/.github/workflows/4.14-clang-13.yml
@@ -41,7 +41,9 @@ jobs:
       LLVM_VERSION: 13
       BOOT: 1
       CONFIG: multi_v7_defconfig
-    container: ghcr.io/clangbuiltlinux/qemu
+    container:
+      image: ghcr.io/clangbuiltlinux/qemu
+      options: --ipc=host
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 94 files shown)


nickdesaulniers (approved):
oh wow, thanks for the fix! :cowboy_hat_face: :moon:

Issue #358, opened 2022-04-27
Test powerpc64le with the integrated assembler on clang-14


nathanchance:
With https://github.com/llvm/llvm-project/commit/33504b3bbe10d5d4caae13efcb99bd159c126070 now merged into `release/14.x`, I believe we can apply #326 to LLVM 14. I will try to test this tomorrow.

PR #360, opened 2022-04-28
Build powerpc64le configs with LLVM=1 with LLVM 14

nathanchance:
Same as commit 4eec5c4 ("generator.yml: Build powerpc64le configs with
LLVM=1 with LLVM ToT") but for LLVM 14, as the patch that allows us to
build with the integrated assembler has been merged into release/14.x:

https://github.com/llvm/llvm-project/commit/33504b3bbe10d5d4caae13efcb99bd159c126070

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/358


nickdesaulniers (approved):
full speed ahead! Thanks for the patch!
:shopping: :coat:

PR #361, opened 2022-04-28
boot-utils: Upgrade for implementation defined pointer authentication

nathanchance:
This should make booting certain arm64 kernels quicker.


nathanchance:
Ah, I forgot this also brings in https://github.com/ClangBuiltLinux/boot-utils/commit/4dd7e0b80e6e9274e8e82762f09b0290c03194ce, which we will need when Arch Linux upgrades to QEMU 7.0.0.

nickdesaulniers (approved):
Thanks for the patch! 📦 📃

PR #364, opened 2022-05-12
Fix ChromeOS runs

nathanchance:
Currently, `generate_tuxsuite.py` and `generate_workflow.py` assume we
always generate a set of defconfig builds. However, it checks this by
looking for "defconfig" in the config, which is not true for the
ChromeOS configs. As a result, we generate an empty list of defconfigs,
which tuxsuite does not like.

To fix this, consider ChromeOS configurations defconfigs, as those are
the default configurations for the ChromeOS trees; they just do not live
in the arch folder because they are split up for the different boards
that are supported by ChromeOS.

Link: https://github.com/ClangBuiltLinux/continuous-integration2/runs/6400686892?check_suite_focus=true

### DIFF
--- a/.github/workflows/chromeos-5.10-clang-12.yml
+++ b/.github/workflows/chromeos-5.10-clang-12.yml
@@ -32,25 +32,9 @@ jobs:
         path: builds.json
         name: output_artifact_defconfigs
         if-no-files-found: error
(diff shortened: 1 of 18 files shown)


nickdesaulniers (approved):
LGTM; thanks! 🇫🇴 🏴‍☠️

PR #366, opened 2022-05-17
Disable ChromeOS jobs temporarily

nathanchance:
TuxSuite is having issues with the size of the ChromeOS repo and they
requested that these jobs stop running until they are able to fix it.

Link: https://gitlab.com/Linaro/tuxsuite/-/issues/159#note_949831351


nickdesaulniers (approved):
interesting that these use more memory than our LTO builds. That doesn't sound right. :shrug:

nathanchance:
As far as I understand it, it is the actual repo clone (or whatever TuxSuite does internally) that is failing, not the kernel build. From the logs, it doesn't even look like it gets to the build phase, it fails during provisioning. Thanks for the review!

Issue #363, opened 2022-05-11
test s390 with integrated as


nickdesaulniers:
https://lore.kernel.org/lkml/20220511120532.2228616-1-hca@linux.ibm.com/

similar to #358 integrated assembler support is in the works for s390. When that lands, we should test it

PR #367, opened 2022-05-18
Enable the integrated assembler for s390 on -next and mainline

nathanchance:
The s390 folks are requiring LLVM 14 and newer so that the integrated
assembler can be used. Remove these builds for -next and mainline, as
they won't be able to start.

The s390 folks have updated their asm to work with the integrated
assembler. Enable it for s390 builds on -next and mainline.

We cannot switch to LLVM=1 because ld.lld and llvm-objcopy do not
support s390:

https://github.com/ClangBuiltLinux/linux/issues/1524
https://github.com/ClangBuiltLinux/linux/issues/1530

For now, we will just use `CC=clang LLVM_IAS=1`.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/363
Link: https://git.kernel.org/s390/c/8218827b73c6e41029438a2d3cc573286beee914


nickdesaulniers (commented):
have you tested that LLVM_IAS=1 works today for s390? I thought @hcahca 's kernel patches were pending, and would depend further on
https://reviews.llvm.org/D125982
https://github.com/llvm/llvm-project/issues/55411#issuecomment-1131891178

nathanchance:
I’ll check my build logs from last night but the patches were in `next-20220518`:

https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/log/arch/s390?h=next-20220518

nickdesaulniers:
Maybe @hcahca can clarify, will either:
1. https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/arch/s390?h=next-20220518&id=e9953b729b789c0e2984859e3b2170b7fa8520d5 land, with the requirement of clang-14, or
2. https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/arch/s390?h=next-20220518&id=e9953b729b789c0e2984859e3b2170b7fa8520d5 will be dropped once https://reviews.llvm.org/D125982 lands, with a requirement of clang-15?

hcahca:
> Maybe @hcahca can clarify, will either:
>     1. https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/arch/s390?h=next-20220518&id=e9953b729b789c0e2984859e3b2170b7fa8520d5 land, with the requirement of clang-14, or

We will go this route (option 1). As soon as clang-15 is released we might increase the required clang version once again - just to avoid that somebody by accident adds new code which otherwise would be silently miscompiled with clang-14.

nickdesaulniers (approved):
looks great! Thanks @nathanchance and @hcahca . :fireworks: :partying_face:

PR #368, opened 2022-05-24
Add objtool patch for stable and 5.15

nathanchance:
This objtool patch is now merged into mainline and it is tagged for
stable but stable patches that are merged during the merge window tend
to have a minor delay getting into stable due to the chaos of the merge
window.

Add this patch so that the builds can go back to green (barring any LLVM
regressions...).


nathanchance:
No, small conflict due to lack of [a bug fix](https://git.kernel.org/linus/988f01683c7f2bf9f8fe2bae1cf4010fcd1baaf5) because the [patch that it fixes](https://git.kernel.org/linus/db2b0c5d7b6f19b3c2cab08c531b65342eb5252b) is not in 5.15. I will watch for the stable "patch failed to apply" email and handle it if nobody else does.

PR #369, opened 2022-05-24
Upgrade stable anchor to 5.18

nathanchance:
Linux 5.18 is the latest stable release, mark it as such.

Copy all the mainline builds so that our coverage matches all the new
development that has happened from 5.17 to 5.18.

### DIFF
--- a/.github/workflows/stable-clang-11.yml
+++ b/.github/workflows/stable-clang-11.yml
@@ -431,6 +431,27 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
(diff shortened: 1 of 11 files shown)


nickdesaulniers (approved):
thanks for the patch! next time try to break these up further into distinct commits. Disabling arm allmodconfig and upgrading ppc could have been distinct commits.  Doesn't really matter though...
🥮 🧇

nickdesaulniers:
> I didn't really see much of a point because all I was doing was copying the mainline builds where those changes already happened and doing s/mainline/stable/g.

Maybe it's just the way the diff is generated/displayed then.  Sometimes `git diff` shows confusing changes based on context.

PR #370, opened 2022-05-25
Disable arm64 CFI builds on mainline for clang-{12,13}

nathanchance:
The patch mentioned in commit 0060c13 ("generator.yml: Disable arm64 CFI
builds on -next for clang-{12,13}") is now in mainline.


nickdesaulniers (approved):
thanks for the patch! 🪨 🚀

PR #371, opened 2022-05-26
patches: 5.15: Fix objtool patch

nathanchance:
I didn't realize until after I had made the initial commit that the
patch did not apply cleanly to 5.15 and I never updated this copy to
match. Do so now.


nickdesaulniers (approved):
thanks! You might end up providing this to stable once they find out this fails to apply, too. :fearful: :face_exhaling:

PR #374, opened 2022-06-03
patches: mainline: Drop a RISC-V patch

nathanchance:
It has been merged into mainline:

https://git.kernel.org/linus/61114e734ccb804bc12561ab4020745e02c468c2

Unfortunately, the other one has not made it yet so it still needs to be
applied.


nathanchance:
-next builds will have this patch when they run on Monday so I amended the change to include dropping that.

nathanchance:
Checks passed, merging to keep the builds running through the weekend.

PR #375, opened 2022-06-09
Drop objtool patch on stable and 5.15

nathanchance:
It has been merged into these trees.


nickdesaulniers (approved):
LGTM; thanks! :on: :mechanic:

PR #376, opened 2022-06-10
Re-enable ChromeOS jobs

nathanchance:
The issues on the TuxSuite side should be resolved: https://gitlab.com/Linaro/tuxsuite/-/issues/162


nickdesaulniers (approved):
thanks for circling back, I had forgotten that we disabled these! 🐗 👞

PR #378, opened 2022-06-13
patches: mainline: Add a fix for ARCH=arm allmodconfig

nathanchance:
This is breaking our builds for ARCH=arm allmodconfig. Add the patch
even though it is not related to LLVM so that we can catch other
breakage easier.


nickdesaulniers (approved):
LGTM; thanks!

PR #381, opened 2022-06-23
Disable ChromeOS jobs again

nathanchance:
This has regressed again: https://gitlab.com/Linaro/tuxsuite/-/issues/162


nickdesaulniers:
Rather than delete everything, how about simply leaving the badges on the readme, leaving the config url stuff in the generator, then just commenting out the individual builds with an added comment which is a link to the tuxsuite gitlab issue?

nathanchance:
Will commenting out all the builds work? We'll basically generate empty TuxSuite and GitHub Actions files, which I don't think will work (I suspect at the very least, TuxSuite will error that there are no builds in the build set).

I have to at least delete the GitHub Actions files and at that point, the badges become dead links, which is why I removed them. Probably should have called that out in the commit message...

nickdesaulniers (approved):

PR #389, opened 2022-08-01
Bump llvm_tot anchor and LLVM_TOT_VERSION

nathanchance:
This allows `scripts/generate.sh to` continue to work.

NOTE: We are not actually using clang-16 yet as apt.llvm.org has not
      been updated since the major version was changed on the main
      branch upstream.


nickdesaulniers (approved):
LGTM; thanks for the patch! 🚋 :octocat:

PR #390, opened 2022-08-02
README: Fix badges after #389

nathanchance:
Generated with `scripts/markdown-badges.sh`.


nickdesaulniers (approved):
thanks! glad this is automated now! :dart: :bullettrain_front:

Issue #387, opened 2022-07-27
Add support for LLVM 15


nathanchance:
The `release/15.x` branch has been cut meaning `clang-nightly` will now be `clang-16` and there will eventually be `clang-15` through [apt.llvm.org](https://apt.llvm.org/). The transition plan:

- [x] Wait for LLVM 15 binaries on [apt.llvm.org](https://apt.llvm.org/)
- [x] Report availability of LLVM 15 binaries on [apt.llvm.org](https://apt.llvm.org/) to [TuxMake issue tracker](https://gitlab.com/Linaro/tuxmake) ([done](https://gitlab.com/Linaro/tuxmake/-/issues/192))
- [x] Wait for TuxMake to provide `clang-15` images ([done](https://gitlab.com/Linaro/tuxmake/-/merge_requests/277))
- [x] Wait for TuxSuite to allow `clang-15` to be specified ([done](https://gitlab.com/Linaro/tuxsuite/-/merge_requests/258))
- [ ] Update `generator.yml` to provide `llvm_14` anchor and copy all `llvm_latest` builds to it
- [ ] Update `generator.yml` to shift `llvm_latest` to 15 and copy all `llvm_tot` builds to it

nathanchance:
`clang-15` is now available on [apt.llvm.org](https://apt.llvm.org] (thanks @sylvestre!) so I have [requested it as an option](https://gitlab.com/Linaro/tuxmake/-/issues/192) for `tuxmake`.

sylvestre:
Welcome ;)

nathanchance:
TuxMake is done: https://gitlab.com/Linaro/tuxmake/-/merge_requests/277

I have open an issue for TuxSuite: https://gitlab.com/Linaro/tuxsuite/-/issues/171

PR #394, opened 2022-08-08
Fix mainline and next builds (August 8, 2022)

nathanchance:
See the individual patches for the full details.


nickdesaulniers (approved):
Thanks! :sun_behind_large_cloud: :tr: 

I saw other issues with i386 defconfig today, can you please check if more patches are needed there for mainline?

nathanchance:
> I saw other issues with i386 defconfig today, can you please check if more patches are needed there for mainline?

Sure. Latest build is from seven hours ago and I only see that one warning:

https://github.com/ClangBuiltLinux/continuous-integration2/runs/7728605036?check_suite_focus=true#step:5:94

However, it looks like our builds of clang are ancient, July 19th, so maybe you're hitting a clang-16 regression? TuxMake images are failing:

https://gitlab.com/Linaro/tuxmake/-/jobs/2825093556

Anders submitted a PR for `clang-15` as part of #387, I commented on it with the diff that should fix those images:

https://gitlab.com/Linaro/tuxmake/-/merge_requests/277

PR #396, opened 2022-08-09
patches: next: Drop rtc patch

nathanchance:
It is available in next-20220809.


nickdesaulniers (approved):
LGTM; thanks! :owl: :mobile_phone_off:

PR #398, opened 2022-08-10
Fix -tip (August 10, 2022)

nathanchance:
The tip tree has inherited the current issues with mainline. Apply the
same patches to keep the trees green while they get fixed.


nickdesaulniers (approved):
LGTM; thanks! :koala: :mosquito:

PR #399, opened 2022-08-10
patches: mainline: Add a couple in-flight -Wformat fixes

nathanchance:
The patch to enable `-Wformat` was merged before these fixes were
available in mainline. Apply them until they are merged so that we do
not have to disable `-Werror` for x86_64.


nickdesaulniers (approved):
thanks; we'll get there! :champagne: :rock: 

cc @JustinStitt

PR #401, opened 2022-08-10
Clean up generator.yml (August 10th, 2022)

nathanchance:
A small series to enable certain build and boot tests and remove some commented out builds that will never be re-enabled. See the individual commits for the details.

### DIFF
--- a/.github/workflows/5.15-clang-12.yml
+++ b/.github/workflows/5.15-clang-12.yml
@@ -410,14 +410,14 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _6a30520d164c342e0191a1a2f5c3ae67:
+  _f19fa50ec541ef4ddce215421e645da8:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


nickdesaulniers (approved):
thanks for revisiting these.  I think besides the comments, have the link to the issue in the commit message when there are disabled will provide a link back to the PR from the issue itself.  The hard part is then when we close an issue, we need to check if there's CI changes that disabled anything related to the issue that can now be re-enabled.

Maybe there's a better way...
Either way, thanks for the patch!
🛢️ 🎥

nathanchance:
> The hard part is then when we close an issue, we need to check if there's CI changes that disabled anything related to the issue that can now be re-enabled.

Right and that is occasionally difficult when a change gets backported to stable and the commits list becomes gigantic...

> Maybe there's a better way...

I think it is probably worth creating two separate issues: one in the main repo that tracks getting the bug fixed in upstream Linux or LLVM and one in this repo for reverting any workarounds that we have to add for said issue. I'll take a peak at the list tomorrow and file issues for them.

Thanks for the review!

PR #407, opened 2022-08-12
patches: mainline: Drop -Wformat patches

nathanchance:
Unfortunately, Linus reverted the -Wformat enablement patch so these are
no longer necessary to keep the CI green. Drop them to reduce our
maintenance overhead.


nickdesaulniers (approved):
LGTM; thanks! 🚚 🇮🇲

PR #410, opened 2022-08-22
patches: Add pending -Wbitfield-constant-conversion mchp-spdiftx.c patch

nathanchance:
This patch is needed to avoid disabling -Werror for x86_64 with clang-16
on mainline, stable, and 5.15. The patch is already accepted so it is in
-next but it is applied to a 6.1 branch so it won't be in mainline for a
couple of months.


nickdesaulniers:
It sounds like we're going to need this fixed in mainline sooner than 6.1; ideally the 6.0-rcX releases. Can you ping the maintainer and ask that they fast track this fix?

nickdesaulniers (approved):
LGTM; thanks 🟠 📙

nathanchance:
> It sounds like we're going to need this fixed in mainline sooner than 6.1; ideally the 6.0-rcX releases. Can you ping the maintainer and ask that they fast track this fix?

Sure, I guess it doesn't hurt to ask, although it is already applied so I guess we would just end up with two copies of the same change, since the ASoC tree doesn't rebase. I can ping the patch tomorrow morning to see what Mark thinks. Thanks for the review!

PR #411, opened 2022-08-23
patches: Add mchp-spdiftx.c to -tip trees

nathanchance:
This patch is needed in the -tip tree as well.


nickdesaulniers (approved):
LGTM; thanks! 🕉️ 🦉

PR #412, opened 2022-08-24
Stop disabling CONFIG_WERROR for arm64

nathanchance:
With the `-Wbitfield-constant-conversion` warning patched up, all the trees that have CONFIG_WERROR build cleanly with allmodconfig and friends for arm64 (aside from -next, which is always a moving target).

Stop disabling it so that we notice build regressions quicker.


nickdesaulniers (approved):
LGTM; thanks! :onion: :es:

PR #414, opened 2022-08-25
patches: arm64: Add a couple more patches to fix warnings

nathanchance:
The arm64 tree has not been moved forward to a 6.0-rc release yet
because no patches for 6.1 have been picked up yet. As a result, we need
to apply these fixes which were merged into mainline during the 5.19 and
6.0 development cycles to avoid errors due to -Werror.


nickdesaulniers (approved):
LGTM; thanks! :owl: :om:

PR #415, opened 2022-08-29
Revert "patches: arm64: Add a couple more patches to fix warnings"

nathanchance:
This reverts commit ee5cf0173dfb54c35d9cf87729ca247f6be82393.

The arm64 for-next/core branch has been bumped to 6.0-rc3, which
includes these patches.

PR #416, opened 2022-08-31
patches: arm64: Update mchp-spdiftx.c patch

nathanchance:
I forgot to update this patch as part of #415. The previous version was
a backport targeting 5.19, this version applies to current mainline.


nickdesaulniers (approved):
LGTM; thanks! :mobile_phone_off: :moon:

PR #417, opened 2022-08-31
Disable CONFIG_WERROR for Hexagon

nathanchance:
A patch in -next introduces the same NULL pointer arithmetic warnings
that we see on s390 to hexagon, which are fatal with -Werror. To keep
the build green, disable CONFIG_WERROR.

Link: https://git.kernel.org/broonie/regmap/c/81c0386c1376da54f05d6916936db5220df9f97d
Link: https://github.com/ClangBuiltLinux/linux/issues/1285


nickdesaulniers (approved):
LGTM; thanks! :waffle: :mask:

PR #421, opened 2022-09-06
Disable CONFIG_PSERIES_PLPKS for Fedora's ppc64le configuration

nathanchance:
This is a new driver in 6.0 that Fedora recently enabled but it causes a
crash during boot, which is not clang specific. Disable it to keep
mainline and -next green while a fix can be found.

Link: https://src.fedoraproject.org/rpms/kernel/c/a73f6858a2cbd16bbcc6d305d6c43aab6f59d0b1
Link: https://lore.kernel.org/Yxe06fbq18Wv9y3W@dev-arch.thelio-3990X/

### DIFF
--- a/.github/workflows/5.15-clang-11.yml
+++ b/.github/workflows/5.15-clang-11.yml
@@ -678,15 +678,15 @@ jobs:
         name: output_artifact_distribution_configs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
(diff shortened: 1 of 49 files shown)


nickdesaulniers (approved):
LGTM; thanks! 🦉 🌔 

Thanks for bisecting and reporting upstream, too!

PR #423, opened 2022-09-08
Disable CONFIG_KCSAN builds for tip of tree LLVM

nathanchance:
Due to the issue linked. Once the kernel side change is finalized, it
can be applied and these builds can be re-enabled.


nickdesaulniers (approved):
LGTM; thanks! :office: :low_brightness:

PR #427, opened 2022-09-10
patches: tip: Drop mchp-spdiftx patch

nathanchance:
tip has been synced with latest mainline, which includes this patch.


nickdesaulniers (approved):
LGTM; thanks! 🚋 🥃

PR #429, opened 2022-09-15
Drop stable tree patches (September 15th, 2022)

nathanchance:
This patch is available in 5.19.9 and 5.15.68.


nickdesaulniers (approved):
LGTM; thanks! :truck: :tamale:

Issue #431, opened 2022-09-19
whitespace bug in generator


nickdesaulniers:
https://github.com/ClangBuiltLinux/continuous-integration2/pull/428#discussion_r974532069

PR #430, opened 2022-09-16
Drop mainline patches (September 16th, 2022)

nathanchance:
The series that properly fixes this has been merged into mainline:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/?h=41012d715d5d7b9751ae84b8fb255e404ac9c5d0

PR #432, opened 2022-09-19
Fix location of '--patch-series' in tuxsuite commands

nathanchance:
The patch series flag should be before the TuxSuite YAML file in the
command and the space should be after the command, rather than before
it, so that the space is between the argument to the flag and the
TuxSuite YAML file.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/431


nathanchance:
I left out the mainline updates due to #430, which is causing presubmit checks to fail here. Once #430 is merged, I can rebase this on that and everything should be green.

generate_workflow.py
```diff
@@ -105,7 +105,7 @@ def tuxsuite_setups(job_name, tuxsuite_yml, repo, ref):
                 },
                 {
                     "name": "tuxsuite",
-                    "run": "tuxsuite plan --git-repo {} --git-ref {} --job-name {} --json-out builds.json {}{} || true".format(repo, ref, job_name, tuxsuite_yml, patch_series)
+                    "run": "tuxsuite plan --git-repo {} --git-ref {} --job-name {} --json-out builds.json {}{} || true".format(repo, ref, job_name, patch_series, tuxsuite_yml)
```
nickdesaulniers, line 108: add whitespace between the `{}{}`

utils.py
```diff
@@ -174,5 +174,5 @@ def patch_series_flag(tree):
     patches_folder = pathlib.Path("patches").joinpath(tree)
     patch_files = glob.glob(
         ci_folder.joinpath(patches_folder).as_posix() + "/*.patch")
-    return " --patch-series {}".format(
+    return "--patch-series {} ".format(
```
nickdesaulniers, line 177: then remove the trailing whitespace

generate_workflow.py
nathanchance: I did it this way to avoid two spaces when there is no patch series but since it is purely cosmetic, I can just do what you have requested.

nathanchance: Done.

utils.py
nathanchance: Done.

nickdesaulniers (approved):
ah, I see now. So it's a tradeoff between position independence and extra whitespace when patch series are unnecessary.  In that case, I don't feel strongly either way.

nathanchance:
In that case, I'll drop e0f97f41a4b42ae3adae7378817bc47a2b72483c since it adds more churn to the generated files and the final result looks odd to me. We are not shuffling the arguments to `tuxsuite` often (if at all) so I don't think the position independence issue is that big of a deal. Thanks for the review!

PR #433, opened 2022-09-20
Apply drm/amd/display patches for dml314 to mainline

nathanchance:
These patches are needed to keep CONFIG_WERROR enabled for x86_64. They
should be in mainline soon, as they have been accepted.


nickdesaulniers (approved):
LGTM; thanks! :tangerine: :movie_camera:

Issue #435, opened 2022-09-20
wire up alpine's kconfigs


nickdesaulniers:
I think https://git.alpinelinux.org/aports/tree/community/linux-edge?h=master might be the link. Or LTS rather than edge:
https://git.alpinelinux.org/aports/tree/main/linux-lts?h=master

https://wiki.alpinelinux.org/wiki/Kernels -> "several kernels" -> "linux-edge" or "linux-lts" -> "Git repository"

PR #438, opened 2022-09-22
Add Alpine Linux configurations

nathanchance:
Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/435


nathanchance:
> I wonder if distro-specific configs based on 5.19 are an issue for older branches of LTS?

They shouldn't be, other than warnings that certain configuration values may not be the correct type (like the symbol being `tristate` in mainline but `bool` in stable). Kconfig can already handle that in a non-fatal manner so it should be fine. Worst case, we just disable building these configurations on older kernels.

> Also, consider reordering the last patch to go first; though we never really need to bisect this repo...

Ack, good point, will do.

PR #439, opened 2022-09-22
utils.py: Handle OpenSUSE riscv64 configuration in get_cbl_name()

nathanchance:
Otherwise, `check_logs.py` will fail.


nickdesaulniers (approved):
LGTM; thanks! :bee: :muscle:

Issue #424, opened 2022-09-08
re-enable kcsan


nickdesaulniers:
https://github.com/ClangBuiltLinux/continuous-integration2/pull/423/commits/59678cbbae1a81d34e7d1cbc970c5fecda40bcbe

nickdesaulniers:
probably -next has @melver 's patches

nathanchance:
No, they are not in -next yet. Looks like Paul updated his `rcu/next` branch yesterday morning:

https://git.kernel.org/pub/scm/infra/transparency-logs/gitolite/git/1.git/tree/m?id=c544c2ac7f5c859baa71da8476e0a5a0e92be270

but -next has not had a release with that updated branch yet, as there was not one today :

https://lore.kernel.org/20220921231304.5a5d0624@canb.auug.org.au/

Should be there in the next -next release though.

PR #440, opened 2022-09-23
Enable x86_64 KCSAN build on -next

nathanchance:
The patch that we need for ToT LLVM is now available:

https://git.kernel.org/next/linux-next/c/fdc1eea7a8473ca30ef8d180e3f99a4955155c7f


nickdesaulniers (approved):
LGTM; thanks! :coconut: :wolf: 

#424

PR #442, opened 2022-09-23
generate_workflow.py: Do not run TuxSuite workflows on pushes to the main branch

nathanchance:
We run all TuxSuite workflows on a schedule so they will eventually run
with whatever fixes are being pushed to the main branch. Automatically
running TuxSuite workflows when they are modified on the main branch is
potentially expensive and could jack up the schedule set in place by the
crons. Allow explicit presubmit runs via "presubmit/" prefixed branches
but do not run the workflows when pull requests are merged to main.

This matches the current reality when I merge pull requests, which is to
just add '[skip actions]' to the merge message so that they do not start
building.

### DIFF
--- a/.github/workflows/4.14-clang-13.yml
+++ b/.github/workflows/4.14-clang-13.yml
@@ -5,7 +5,6 @@ name: 4.14 (clang-13)
 'on':
   push:
     branches:
-    - main
     - presubmit/*
     paths:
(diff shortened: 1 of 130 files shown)


nickdesaulniers (approved):
LGTM; thanks! :turkey: :turtle:

PR #446, opened 2022-09-27
boot-utils: Update to latest main

nathanchance:
To pick up:

6a2833c boot-qemu: Move from 'powernv8' to 'powernv' machine
09fd7a3 boot-utils: Flush print commands to ensure output is properly ordered
a0540ce boot-utils: Remove skiboot
5afe426 Expand version matching


nickdesaulniers (approved):
LGTM; thanks! :waffle: :ireland:

PR #456, opened 2022-10-13
Re-enable CONFIG_FPE_NWFPE for arm all{yes|mod}config

nickdesaulniers:

PR #458, opened 2022-10-17
Fix application of NWFPE patches on Android and stable trees

nathanchance:
I noticed that certain Android and stable trees were failing with the link failure that the patch added by #456 should have fixed. As it turns out, certain patch folders have the wrong name so they were not picked up by the generator. I have fixed this and added a new check in `scripts/check-patches.sh` to catch this in the future.


nickdesaulniers (approved):
What a terrible mistake I made!

Always be looking for opportunities to make mistakes preventable; good job! 🍕 🛢️

nickdesaulniers:
odd, both github UI and git show a funny looking rename:

```
rename from patches/android12-5_4/series
rename to patches/5.15/series
```
which doesn't look right, but they're the same file contents...

PR #460, opened 2022-10-21
CONFIG_KASAN_KUNIT_TEST requires tracing now

nathanchance:
After commit fcfe5ee1f509 ("kasan: switch kunit tests to console
tracepoints") in -next.

To avoid having to reshuffle the configs due to the increased length,
break out the arm64 KASAN configs into a separate section, similar to
chromeos_configs.

Link: https://git.kernel.org/next/linux-next/c/fcfe5ee1f50914cf3acc4fe389d2ce67a0183f41
Link: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/3297390131/jobs/5439007513

### DIFF
--- a/.github/workflows/5.15-clang-11.yml
+++ b/.github/workflows/5.15-clang-11.yml
@@ -242,15 +242,15 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _cb277ab066f41ea2734315fd1f98cf92:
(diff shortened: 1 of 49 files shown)


nathanchance:
Small ping on this. Also, I forgot to notate that this is not needed for the x86 KASAN configs, as ftrace is enabled in defconfig there.

nickdesaulniers (approved):
LGTM; Thanks!

PR #461, opened 2022-10-31
patches: Apply patch for bpf recordmcount issue

nathanchance:
This error is breaking builds in mainline and -next. Apply the patch
that was sent upstream to resolve this.


nickdesaulniers (approved):
LGTM; thanks!

PR #462, opened 2022-10-31
patches: Apply 2120635108b35ecad9c59c8b44f6cbdf4f98214e to arm64 trees

nathanchance:
The arm64 and arm64-fixes trees have not been updated to an 6.1 rc
release yet, which means they do not have commit 2120635108b3
("Makefile.extrawarn: Move -Wcast-function-type-strict to W=1"), which
is needed to avoid build failures with ToT clang.

While we wait for the arm64 trees to be updated, apply that change to
the tree to avoid the build failures with -Werror due to the new
warnings.


nickdesaulniers (approved):
LGTM; thanks!

PR #464, opened 2022-11-01
Increase 'tuxsuite plan' step timeout to eight hours

nathanchance:
By default, GitHub Actions limits each step in a workflow to 360
minutes, which would normally be enough but if an all*config build gets
preempted and has to restart, it is possible for the 'tuxsuite plan'
command to exceed 360 minutes. Increase it by two more hours up to a
total of eight hours to see if that is enough to avoid job timeouts.


nickdesaulniers:
This might need rebasing due to collision with #463.

nathanchance:
> We should see how feasible it is to remove this (or tighten down the timeout) in the future.

Sure, that definitely seems reasonable. That will depend on whether or not TuxSuite could do something about the infrastructure error retries that it does, as that appears to be the reason that we need to do this. Thanks for the review!

PR #467, opened 2022-11-07
patches/next: Update bpf patch

nathanchance:
Commit dbe69b299884 ("bpf: Fix dispatcher patchable function entry to 5
bytes nop") is being reverted in favor of commit c86df29d11df ("bpf:
Convert BPF_DISPATCHER to use static_call() (not ftrace)") so the
current patch no longer applies. Remove it.

Unfortunately, c86df29d11df introduces a new warning that will break
builds with -Werror. Add my patch that fixes that warning.


nickdesaulniers (approved):
2 steps forwards, 1 step back!

Thanks for the patch! :rocket: :moon:

PR #469, opened 2022-11-07
Drop arm64 patches (November 7th, 2022)

nathanchance:
The arm64 has now been updated to Linux 6.1-rc4, which includes both of
these patches. Drop them altogether.


nickdesaulniers (approved):
LGTM; thanks! :train: :tree:

PR #473, opened 2022-11-17
patches: mainline: Drop BPF patch

nathanchance:
dbe69b299884 was reverted in favor of a better fix (c86df29d11dfba2),
which is now in mainline, so this patch is no longer applicable. Drop
it.


nickdesaulniers:
Specifically,
commit 18acb7fac22f ("bpf: Revert ("Fix dispatcher patchable function entry to 5 bytes nop")")
was the revert.

nickdesaulniers (approved):
LGTM; thanks! :trinidad_tobago: :meat_on_bone:

PR #474, opened 2022-11-17
Add more checks for potential TuxSuite problems

nathanchance:
This PR adds some more explicit checks for situations that we have hit in TuxSuite, such as an empty builds.json or unfinished builds, which will help making triaging failures quicker.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -118,7 +118,7 @@ def _read_builds():
         builds = "mock.builds.json"
     try:
         if pathlib.Path(builds).stat().st_size == 0:
-            raise Exception("%s is zero sized?" % (builds))
+            raise Exception(f"{builds} is zero sized?")
         with open(builds) as f:
             builds = json.load(f)
     except FileNotFoundError as e:
--- a/utils.py
+++ b/utils.py
@@ -117,6 +117,8 @@ def _read_builds():
     if os.environ.get("MOCK"):
         builds = "mock.builds.json"
     try:
+        if pathlib.Path(builds).stat().st_size == 0:
+            raise Exception("%s is zero sized?" % (builds))
         with open(builds) as f:
             builds = json.load(f)
     except FileNotFoundError as e:
--- a/check_logs.py
+++ b/check_logs.py
@@ -247,6 +247,9 @@ def run_boot(build):
 
 
 def boot_test(build):
+    if build["result"] == "unknown":
+        print_red("unknown build result, skipping boot")
+        sys.exit(1)
     if build["result"] == "fail":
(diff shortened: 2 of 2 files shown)


utils.py
```diff
@@ -117,6 +117,8 @@ def _read_builds():
     if os.environ.get("MOCK"):
         builds = "mock.builds.json"
     try:
+        if pathlib.Path(builds).stat().st_size == 0:
```
nickdesaulniers, line 120: prefer `os.path.getsize(builds)`

utils.py
```diff
@@ -117,6 +117,8 @@ def _read_builds():
...
+            raise Exception("%s is zero sized?" % (builds))
```
nickdesaulniers, line 121: do you want to use an `f` string here?

utils.py
nathanchance: Sure, although the documentation says that these return equivalent numbers:

https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat
https://docs.python.org/3/library/os.html#os.stat_result.st_size
https://docs.python.org/3/library/os.path.html#os.path.getsize

I prefer `pathlib` because it is easy to reason about but I will change it.

utils.py
nathanchance: Oops, yes! This was the thing that made me ask about the conversions in the first place so this should definitely be updated.

utils.py
nathanchance: a5955b6c8047fe8c3769d35fc7b2126a576c1df7

utils.py
nathanchance: 49e96e7a5b8e72b9cccf8a9b2becca18efd6cc78

nickdesaulniers (approved):
LGTM; thanks! :t-rex: :mosquito:

utils.py
nickdesaulniers: [citation needed]

nathanchance: Not super authoritative but good reads:

https://towardsdatascience.com/dont-use-python-os-library-any-more-when-pathlib-can-do-141fefb6bdb5

https://stackoverflow.com/questions/67112343/pathlib-vs-os-path-join-in-python

https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

nickdesaulniers: TIL (Stupid python stdlib full of dead batteries) ok, feel free to drop a5955b6c8047fe8c3769d35fc7b2126a576c1df7 then

ojeda: @nickdesaulniers The docs say `pathlib` is the high-level interface (objects vs. strings). There is also rationale in the PEP that introduced it.

PR #475, opened 2022-11-17
continuous-integration2: Python updates

nathanchance:
The commit messages should be self explanatory but this PR updates boot-utils to use f-strings (preferred in Python 3.6+) and eliminates as_posix() calls on path objects (possible after Python 3.6+), which cleans up the code a bit.


check_logs.py
```diff
@@ -26,31 +26,28 @@ def _fetch(title, url, dest):
             urllib.request.urlretrieve(url, dest)
             break
         except ConnectionResetError as err:
-            print_yellow('%s download error ("%s"), retrying...' %
-                         (title, str(err)))
+            print_yellow(f'{title} download error ("{str(err)}"), retrying...')
```
nickdesaulniers, line 29: while you're here, mind changing this to double quotes? Let's be consistent throughout the codebase.

check_logs.py
```diff
@@ -26,31 +26,28 @@ def _fetch(title, url, dest):
...
+                    f"{title} download error ({err.code}), retrying...")
                 pass
             else:
-                print_red("%d error trying to download %s" % (err.code, title))
+                print_red(f"{err.code} error trying to download {title}")
                 sys.exit(1)
         except urllib.error.URLError as err:
-            print_yellow('%s download error ("%s"), retrying...' %
-                         (title, str(err)))
+            print_yellow(f'{title} download error ("{str(err)}"), retrying...')
```
nickdesaulniers, line 40: dbl quote

check_logs.py
nathanchance: 5a4758e7e3ec8d7e70afc20ab0c30b1ea95dbd70

check_logs.py
nathanchance: 5a4758e7e3ec8d7e70afc20ab0c30b1ea95dbd70

nickdesaulniers (approved):
LGTM; thanks! :om: :ascension_island:

PR #476, opened 2022-11-18
boot-utils: Update to latest main

nathanchance:
To pull in:

ca70cb3 Eliminate calling as_posix() on path objects
6b30186 Switch to f-strings
c5b96f8 ignore unicode decode errors on macOS
2a94d4a Fix cpu count on macOS.


nickdesaulniers (approved):
LGTM; thanks! :sun_behind_large_cloud: :moon:

PR #477, opened 2022-11-28
Adjust location of raised timeout in GitHub Actions workflow files

nathanchance:
The eight hour timeout needs to be set in the job scope, rather than the
step scope, otherwise the entire job will be killed by its default six
hour timeout.


nickdesaulniers (approved):
LGTM; thanks! Sorry I missed this the first time around.

Issue #478, opened 2022-11-29
add button for check-clang-version


nickdesaulniers:
https://github.com/ClangBuiltLinux/continuous-integration2/actions/workflows/clang-version.yml has been red but we didn't notice because it's not in the grid of buttons we check in the readme or homepage.  Probably doesn't need to be in the grid, but we should have an "indicator light" somewhere.

sylvestre:
My bad:
`clang hasn't been updated in 25 days!`

I have some issues that I don't know how to fix:
https://github.com/llvm/llvm-project/issues/59097

nathanchance:
> My bad: `clang hasn't been updated in 25 days!`

To be fair, the reason for that is not entirely your fault :) TuxMake's images have been failing to build for three weeks now, which was before apt.llvm.org started failing:

https://gitlab.com/Linaro/tuxmake/-/issues/196

sylvestre:
I fixed the issue on my side. we should be up to date now on all branches!

PR #479, opened 2022-12-03
Add "Check clang version" badge to README

nathanchance:
This workflow failing means that the version that we are using is not
staying in sync with tip of tree LLVM, which means we are potentially
losing out on test coverage. This should be more obvious so that it can
be dealt with quicker so add it to the markdown badge generator. It does
not need to be included in the actual table so it is added to the end.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/478


nickdesaulniers (approved):
LGTM; Thanks! :o2: :mobile_phone_off:

PR #481, opened 2022-12-05
patches: android13-5.10: Add upstream series for pahole 1.24 compatability

nathanchance:
android13-5.10 enables BTF generation via pahole. TuxMake's version of
pahole was recently bumped to the latest (1.24) as part of an overall
update to Debian testing, which generates `BTF_KIND_ENUM64` by default,
even though 5.10 does not have support for it, resulting in the
following failure:

```
    FAILED: load BTF from vmlinux: Unknown error -22
```

This is resolved in upstream 5.10 with a series to disable generating
these BTF types but that may take some time to get merged into the
Android trees due to release times lines so apply the series via
tuxsuite for the time being.


nickdesaulniers (approved):
LGTM; thanks! :moon_cake: :weary:

PR #482, opened 2022-12-09
generator.yml: Disable x86_64 allyesconfig builds on -tip

nathanchance:
Due to increased memory pressure from objtool causing timeouts.


nickdesaulniers (approved):
LGTM; thanks :maple_leaf: :hocho:

PR #483, opened 2022-12-09
Add x86 vDSO patch for tip of tree ld.lld default change

nathanchance:
mainline and -next also need this technically but the patch is in the x86/urgent branch so it should make 6.1, which is on Sunday so I do not see a point in adding it; the builds can just fail until then. -next is not getting updated again until Tuesday so similar situation as mainline.


nickdesaulniers:
Will a patch in patches/stable conflict with a patch in patches/5.15?

nickdesaulniers:
> Will a patch in patches/stable conflict with a patch in patches/5.15?

I assume no, but I forget how the patch application works.

nickdesaulniers (approved):
LGTM; thanks :wolf: :lobster:

nathanchance:
> Will a patch in patches/stable conflict with a patch in patches/5.15?

Nope, separate trees. stable is the latest stable release, not latest LTS release, which is still currently 5.15.

nickdesaulniers:
> > Will a patch in patches/stable conflict with a patch in patches/5.15?
> 
> Nope, separate trees. stable is the latest stable release, not latest LTS release, which is still currently 5.15.

ah, right, stable is a floating version.

PR #485, opened 2022-12-12
Update stable anchor to 6.1

nathanchance:
The first commit is something I noticed when updating these builds. It could have been a separate pull but it makes sense in the context of the second commit.


nickdesaulniers (approved):
LGTM; thanks! :cheese: :wine_glass:

PR #487, opened 2022-12-15
Adjust mainline patches (December 15th, 2022)

nathanchance:
This pull request adds two more patches for mainline while removing the current one. Hopefully we will be down to zero patches soon...


nickdesaulniers (approved):
LGTM; thanks! 🕉️ 🦢

PR #488, opened 2022-12-16
Adjust x86_64 allyesconfig builds

nathanchance:
This adjusts mainline and -next to disable x86_64 allyesconfig to match -tip, while turning on x86_64 allyesconfig for -tip, as I do not believe the issue is reproducible there.


nathanchance:
Ping, this might have been overlooked.

nickdesaulniers (approved):
yeah, sorry I missed this! Thanks for the patch!

Let's keep an eye on the results of clang-14 after this lands.

PR #493, opened 2022-12-22
Add fixes to linux-next during break

nathanchance:
Stephen Rothwell is taking a break from updating -next over the
holidays, until January 5th:

https://lore.kernel.org/20221220112148.64f183ce@canb.auug.org.au/

While there is the possibility of updates from Mark Brown, they are not
guaranteed, which would mean that our builds of -next will be red for
quite a bit of time.

To keep the tree green during this time of no updates, apply the patches
that are outstanding to resolve all reported build problems, which will
reduce our failure emails and serve as a good build canary for LLVM tip
of tree updates over the break.

These will need to be removed before the next -next update, which I have
a reminder set to do on January 4th.


nickdesaulniers (approved):
LGTM; thanks! :rock: :roller_skate:

PR #494, opened 2022-12-26
patches: next: Update for next-20221226

nathanchance:
Stephen releases next-20221226, which was unexpected. Drop the patches
that were applied to keep the build green until his return.

Unfortunately, with that update, we picked up two new issues. I have
reported them to the authors upstream so they will hopefully be fixed by
the next -next update.


nathanchance:
Presubmit checks are passing, merging this PR without explicit approval due to people being on holidays.

PR #496, opened 2022-12-30
Rewrite all shell scripts in Python

nathanchance:
This PR replaces all shell scripts with Python equivalents. Even for the smaller scripts, this can result in something that is more readable, safer, and faster. I have tried to comment on things that look tricky but this should be a lot simpler to overstand overall. The biggest change is `generate.py`, which is now able to just call `generate_tuxsuite.py` and `generate_workflow.py` directly through the main functions.


ojeda:
This is great -- I don't know how the performance compares, but looks better to me indeed :)

scripts/check-matrix.py
```diff
@@ -0,0 +1,32 @@
...
+    if (jobs := len(jobs_re.findall(workflow_text))) > GH_LIMIT:
+        overlimit_workflows[workflow.name] = jobs
+
+if overlimit_workflows:
+    print(
+        f"The following workflows have more than {GH_LIMIT} jobs, please reduce the number of jobs or split the matrix up:\n"
+    )
+    for k, v in overlimit_workflows.items():
+        print(f"{k} ({v})")
+    sys.exit(1)
```
nickdesaulniers, line 32: as soon as we find the first overflow, we could print then stop iterating (sys.exit)

nickdesaulniers (approved):
nice! I'm surprised that python is faster than shell (maybe depends on which shell, but w/e). Nice work! Thanks :woman_beard: :weary:

scripts/check-matrix.py
nathanchance: I considered that but then I remembered that I wrote it this way so that we catch ALL workflows that are over the limit, rather than just the first. It would rather annoying to have two workflows over the limit, fix the first one, run the check again, and have it still fail. That probably deserves a comment.

PR #500, opened 2023-01-07
4.9 is now EOL

nathanchance:
Link: https://lore.kernel.org/Y7lbu6%2F0P7Q%2FP3oj@kroah.com/


nickdesaulniers:
Follow up from https://github.com/ClangBuiltLinux/ClangBuiltLinux.github.io/pull/47, we can probably stop the Android 4.9 builds now, too.

PR #505, opened 2023-01-23
Drop mainline patch (January 23, 2023)

nathanchance:
The error fixed by this patch has been resolved through other patches.


nickdesaulniers (approved):
LGTM; thanks! 🆒 🍳

Issue #506, opened 2023-01-25
setup clang-16 branches


nickdesaulniers:
Sounds like clang-16 is now branched: https://github.com/ClangBuiltLinux/linux/issues/1788

meaning ToT clang is now clang-17. We need to setup dedicated branches for clang-16.

nathanchance:
Plan:

- [ ] Update LLVM main anchor and version to 17
- [ ] Wait for LLVM 16 binaries on [apt.llvm.org](https://apt.llvm.org/)
- [ ] Report availability of LLVM 16 binaries on [apt.llvm.org](https://apt.llvm.org/) to [TuxMake issue tracker](https://gitlab.com/Linaro/tuxmake)
- [ ] Wait for TuxMake to provide `clang-16` images
- [ ] Wait for TuxSuite to allow `clang-16` to be specified
- [ ] Update `generator.yml` to provide `llvm_15` anchor and copy all `llvm_latest` builds to it
- [ ] Update `generator.yml` to shift `llvm_latest` to 16 and copy all `llvm_tot` builds to it

sylvestre:
the 16 packages are available on apt.llvm.org I haven't updated the home page yet :)

dileks:
Thanks!

Hmm, `clang-16` both in `llvm-toolchain-16/main` and `llvm-toolchain/main` (snapshot).
Hope, `clang-17` will follow.
```
root# LC_ALL=C apt-cache policy clang-16
clang-16:
  Installed: (none)
  Candidate: 1:16.0.0~++20230125063415+0604154e006e-1~exp1~20230125183531.2
  Version table:
     1:16.0.0~++20230125063415+0604154e006e-1~exp1~20230125183531.2 99
         99 https://apt.llvm.org/unstable llvm-toolchain-16/main amd64 Packages
     1:16~++20230124111854+2a539ee17d8a-1~exp1~20230124232010.1062 99
         99 https://apt.llvm.org/unstable llvm-toolchain/main amd64 Packages
     1:16~++20221111053703+a77a02aa5794-1~exp1 99
         99 https://ftp.debian.org/debian experimental/main amd64 Packages
         99 https://deb.debian.org/debian experimental/main amd64 Packages

root# LC_ALL=C apt-cache policy clang-17
N: Unable to locate package clang-17

[...]

PR #507, opened 2023-01-26
LLVM main is now 17

nathanchance:
While clang-nightly in TuxSuite has not updated to clang-17 just yet, it
should be soon and this is needed so that updating the generated files
works properly.

Part one of https://github.com/ClangBuiltLinux/continuous-integration2/issues/506


nickdesaulniers (approved):
LGTM; thanks! :metal: :heavy_check_mark:

PR #161, opened 2021-07-06
Disable s390 builds for clang 12 and earlier on mainline

nathanchance:

PR #510, opened 2023-02-01
Disable s390 builds with LLVM < 13 on 5.10

nathanchance:
Commit 82d3edb50a11 ("s390/cpum_sf: add READ_ONCE() semantics to compare
and swap loops") was applied to 5.10, where it started causing build
breakage with LLVM 11 and 12:

```
  arch/s390/kernel/perf_cpum_sf.c:1234:3: error: inline asm error: This value type register class is not natively supported!
                  "       cdsg    %[old],%[new],%[ptr]\n"
                  ^
  warning: Compiler has made implicit assumption that TypeSize is not scalable. This may or may not lead to broken code.
```

The construct that is used in the inline assembly was only supported in
LLVM starting with the 13.0.0 release, even though it has been supported
in GCC for a while. This was the catalyst for commit e2bc3e91d91e
("scripts/min-tool-version.sh: Raise minimum clang version to 13.0.0 for
s390").

The min-tool-version.sh infrastructure does not exist in 5.10 and it may
be undesirable to backport it. Do what was done in #150 and #161 and
disable LLVM 11 and 12 builds on stable-5.10. This does not lose us much
coverage, as we were only testing one configuration on these branches.

### DIFF
--- a/.github/workflows/5.10-clang-11.yml
+++ b/.github/workflows/5.10-clang-11.yml
@@ -243,27 +243,6 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _3b0c273122a784488dcfee7a6a9016b9:
-    runs-on: ubuntu-latest
-    needs: kick_tuxsuite_defconfigs
-    name: ARCH=s390 CC=clang LLVM_IAS=0 LLVM_VERSION=11 defconfig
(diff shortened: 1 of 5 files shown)


nathanchance:
Presubmit should be fixed with #511.

nickdesaulniers (approved):
LGTM; thanks! :om: :owl:

PR #511, opened 2023-02-01
Replace Exception with more specific RuntimeError

nathanchance:
Recent versions of pylint complain:

  W0719: Raising too general exception: Exception (broad-exception-raised)

This does not truly matter since we do not expect these exceptions to be
caught but it is a simple change to make the linter happy. Use the more
specific but still accurate RuntimeError exception, which is used for
generic problems at runtime.


nickdesaulniers (approved):
LGTM; thanks! :moon: :new_moon:

PR #512, opened 2023-02-06
Enable KCSAN on stable

nathanchance:
The change for tip of tree LLVM has been backported to both of these
trees now so it is safe to re-enable these.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/424


nathanchance:
Small ping, in case this was missed.

nickdesaulniers (approved):
LGTM; thanks! :watermelon: :lemon:

PR #517, opened 2023-02-15
Add script to locally build TuxSuite configurations

nathanchance:
Sometimes, it may be desirable to perform the same set of builds that
TuxSuite does locally, such as validating that a local tree does not
have any problems.

TuxMake is the backend of TuxSuite and it is written in Python, so we
can import it, parse the TuxSuite YAML files ourselves, then offload all
the builds onto TuxMake.

Add `scripts/build-local.py` to take care of this and add a note of it in the
tuxsuite configurations.


scripts/build-local.py
```diff
@@ -0,0 +1,165 @@
...
+        "tuxmake requires either podman or docker for versioned toolchains but neither could be found on your system!"
+    )
+
+# Combine all jobs into one object for easy iteration
+jobs = {}
+for file_str in args.files:
+    with open(file_str, encoding='utf-8') as file:
+        config = yaml.safe_load(file)
+
+    current_jobs = config['jobs']
```
nickdesaulniers, line 86: `current_jobs` is unused.

scripts/build-local.py
```diff
@@ -0,0 +1,165 @@
...
+        toolchain = build['toolchain']
+
+        cfg_str = '+'.join(kconfig) if isinstance(kconfig, list) else kconfig
+        print(f"I: Building {target_arch} {cfg_str} ({toolchain})... ",
+              end='\n' if args.verbose else '',
+              flush=True)
+
+        if build_dir.exists():
+            shutil.rmtree(build_dir)
+        build_dir.mkdir(exist_ok=True, parents=True)
```
nickdesaulniers, line 122: probably don't need `exist_ok=True` anymore? If it existed, it would have been removed on the line above.  Though perhaps TOCTOU is a concern between the two statements?

scripts/build-local.py
```diff
@@ -0,0 +1,165 @@
...
+                                     output_dir=specific_output_dir,
+                                     quiet=(not args.verbose),
+                                     runtime=runtime,
+                                     tree=tree,
+                                     **build)
+        passed = True
+        for info in result.status.values():
+            passed &= info.passed
+
+        if passed:
```
nickdesaulniers, line 158: I bet you could use `all` here.

```python
if all(info.passed for info in result.status.values):
```

generate_tuxsuite.py
```diff
@@ -36,8 +36,18 @@ def emit_tuxsuite_yml(config, tree, llvm_version):
...
         print("# Invoke tuxsuite via:")
+        patches_flag = patch_series_flag(tree)
         print(
-            f"# $ tuxsuite plan --git-repo {repo} --git-ref {ref} --job-name defconfigs --json-out builds.json {patch_series_flag(tree)}{tuxsuite_yml}"
+            f"# $ tuxsuite plan --git-repo {repo} --git-ref {ref} --job-name defconfigs --json-out builds.json {patches_flag}{tuxsuite_yml}"
+        )
+        print("# Invoke locally via:")
+        print(f"# $ git clone -b {ref} --depth=1 {repo} linux")
+        if patches_flag:
+            patches_folder = patches_flag.strip().rsplit(' ', 1)[-1]
```
nickdesaulniers, line 46: Mind adding a comment along the lines of:

```python
# Input: foo bar baz
# Output: foobarbaz
```

It's hard for me to imagine what these string operations are doing.

nickdesaulniers (commented):
This is a really cool idea. I love how it includes a comment on how to repro locally. Very very nice! 🐂 🪵

scripts/build-local.py
nathanchance: 38b6ca77caa22c260704538979317f7a3cb016b9

scripts/build-local.py
nathanchance: Yeah, this was paranoia over `shutil.rmtree()` failing but I did not specify `ignore_errors=True` for that call so if it fails, an exception will be raised. I think I will just remove the `exist_ok=True` from the `mkdir()` call. Alternatively, I could do:

```
shutil.rmtree(build_dir, ignore_errors=True)
build_dir.mkdir(exist_ok=True, parents=True)
```

but I think removing `exist_ok=True` is a bit better. Done: e3a1493447355709daeb9bae4f34c9987a40446b

scripts/build-local.py
nathanchance: Thanks, that looks much better! 4cb798b007df22ce470547e58824029f1e73cacf

generate_tuxsuite.py
nathanchance: Sure thing: 849dfd4916a92d2d2281bb5682b1348f4605fc67

generate_tuxsuite.py
```diff
@@ -43,6 +43,8 @@ def emit_tuxsuite_yml(config, tree, llvm_version):
         print("# Invoke locally via:")
         print(f"# $ git clone -b {ref} --depth=1 {repo} linux")
         if patches_flag:
+            # Input: '--patch-series ... '
+            # Output: '...'
             patches_folder = patches_flag.strip().rsplit(' ', 1)[-1]
```
nickdesaulniers, line 48: does
```diff
-patches_folder = patches_flag.strip().rsplit(' ', 1)[-1]
+patches_folder = patches_flag.split(' ')[1]
```
work here?

nathanchance: Sure, I do not see why not, I can change it to that if you prefer.

nickdesaulniers: yes please

nathanchance: Done: e58a2daf8e53386c418025da7ac1c8ad7be6b9e6

nickdesaulniers (approved):
looks great, nice work! :arrow_lower_left: :eight:

nathanchance:
Thanks a lot for the review as always!

Issue #308, opened 2022-02-17
front end caching


nickdesaulniers:
- [ ] can we propagate info between builds?
  - [ ] propagate HEAD shas for the kernel tree
  - [ ] propagate HEAD shas or version in for LLVM (maybe hardcode for released versions)
  - [ ] propagate prior run build status (red/green)
- [ ] at the start of a workflow, refetch SHAs
- [ ] if no change to SHAs, set exit status to prior build status
- [ ] rewrite latest SHAs for propagation

nathanchance:
> can we propagate info between builds?

I think so. There are plenty of actions that will automatically commit to the repository:

https://github.com/devops-infra/action-commit-push
https://github.com/stefanzweifel/git-auto-commit-action
https://github.com/EndBug/add-and-commit

I do not see a way to do this without committing to the repository, we cannot access artifacts across runs (["Note: You can only download artifacts in a workflow that were uploaded during the same workflow run."](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#downloading-artifacts-during-a-workflow-run)) and I do not think we can write environment variables (at least, I have not found a way to in their documentation).

We could potentially write JSON or YAML with the compiler version, kernel version, and build status from the previous run then parse that when a new run starts.

Some initial thoughts:

* Will we run into issues where multiple workflows attempt to update their status files at the same time (`git push` will fail)? We could potentially run `git pull -r` before pushing? It looks like the last action I linked above has support for this.
* To get the current compiler version, we will need to use the TuxMake container, like we do in [`clang-version.yml`](https://github.com/ClangBuiltLinux/continuous-integration2/blob/15409e8d48ebe49e6908f406272a51c0988c2acc/.github/workflows/clang-version.yml).
* `git ls-remote <REPO> <BRANCH> | awk '{print $1}'` to get latest SHA without cloning:
  ```
  $ git ls-remote https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git linux-4.9.y | awk '{print $1}'
  9279031d74f8fe8760ce32ac527bc4658b578926
  ```

nathanchance:
Talking with @kees about this during our weekly meeting, we think this is doable by giving every workflow file its own branch and information file, which eliminates the concern I had over multiple workflows updating one branch.

Rough idea for a `do-we-need-to-run.py` (or whatever we want to do):
- Does the workflow's branch exist in the repository?
[...]

PR #519, opened 2023-02-22
Add ruff.toml and clean up warnings

nathanchance:
[`ruff`](https://github.com/charliermarsh/ruff) is a fast Python linter written in Rust. Add a `ruff.toml` file to turn on several classes of warnings that are useful and fix them up.

PR #522, opened 2023-02-26
Initial stab at front end cache checks

nathanchance:
NOTE: This is currently based on #519 for ease of development, I will rebase this on main once that PR is merged.

There are times where neither the kernel revision nor the compiler have
changed since the last run, which means there is little point running the
build because it is unlikely anything would change (except due to
flakiness with GitHub Actions, more on that later). While tuxsuite does
have compiler caching enabled, it still relies on spinning up all the
build machines, letting the cache work, then firing off the boot tests,
which can be expensive.

By caching the compiler string and the kernel revision, it is possible
to avoid even spinning up the tuxsuite jobs if nothing has changed.
`should_run.py` takes care of this by exiting:

    * 0 if the build should run (something changed or there was no
      previous results)
    * 1 for any internal assertion failure
    * 2 if nothing changed from the previous run

If there is any internal assertion failure, the cache step fails,
failing the whole workflow, so that those situations can be properly
dealt with. The script should never do this, but it is possible there
are things I have not considered happening yet.

To avoid contention with pushing and pulling, each `should_run.py` job
gets its own branch. While this will result in a lot of branches, they
should not cause too many issues because they will just contain the JSON
file with the last run info.

`should_run.py` does not try to account for flakiness on the GitHub
Actions or TuxSuite side, so it is possible that a previous build will
fail due to flakiness and not be retried on the next cron if nothing
changes since then. To attempt to account for a situation where we
re-run a known flaky build, the script gets out of the way when the
GitHub Actions event is `workflow_dispatch`, meaning a workflow was
manually run. Additionally, if the previous run was only flaky during
the QEMU boots (rather than during the TuxSuite stage), they can
generally just be re-run right away, since the kernels do not need to be
rebuilt. I do not think this will happen too often but if it does, we
can try to come up with a better heuristic.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/308

### DIFF
--- a/.github/workflows/4.14-clang-13.yml
+++ b/.github/workflows/4.14-clang-13.yml
@@ -16,8 +16,32 @@ name: 4.14 (clang-13)
   workflow_dispatch: null
 permissions: read-all
 jobs:
+  cache_check_defconfigs:
+    name: cache check (defconfigs)
+    runs-on: ubuntu-latest
+    container: tuxmake/clang-13
+    outputs:
+      should_run: ${{ steps.should_run.outputs.should_run }}
+    permissions: write-all
+    steps:
+    - uses: actions/checkout@v3
+    - name: Should build run?
+      id: should_run
+      run: |
+        if python3 should_run.py || { ret=$?; ( exit $ret ) }; then
+          echo "should_run=true" >>$GITHUB_OUTPUT
+        else
+            case $ret in
+              2) echo "should_run=false" >>$GITHUB_OUTPUT ;;
+              *) exit 1 ;;
+            esac
+        fi
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   kick_tuxsuite_defconfigs:
     name: TuxSuite (defconfigs)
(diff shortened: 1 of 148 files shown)


nathanchance:
> We might be able to use `docker manifest inspect tuxmake/clang-nightly` to check whether the docker container has been updated.

Right, this would save us the pull to check the image. As we discussed, that might result in cache misses compared to the compiler version string, as `tuxmake` rebuilds the images monthly, meaning the digest will change even if the compiler version has not. I think that is okay for a potentially simpler implementation.

> If it has, we'll have to see if tuxmake is putting the same executable in docker (hopefully not).

If I am parsing this correctly, yes, I believe it is  possible that the digest of the image changes while the compiler does not (think old `apt.llvm.org` versions that are not seeing anymore updates).

We could use `docker manifest inspect` to see if the image has changed then pull the image and check its compiler if it has?

JustinStitt:
superseded by #664.

@nathanchance safe to close?

nathanchance:
Yup!

PR #523, opened 2023-02-27
Add script to print the number of builds done per week per workflow

nathanchance:


nickdesaulniers:
Mind pulling this out into its own tool? I think that would help us extend it in the future and keep it focused on just estimations.

nathanchance:
> Mind pulling this out into its own tool? I think that would help us extend it in the future and keep it focused on just estimations.

Done. Additionally, as requested, this relies on `generator.yml`, rather than the workflow files directly. As it turns out, I had a counting bug in the previous version where the `tuxsuite` jobs were getting counted as builds, so this avoids that.

Unfortunately, I cannot update the branch name without closing the pull request (I believe...), so I'll leave that be for now.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,57 @@
...
+
+builds_per_tree = {}
+for tree in config['tree_schedules']:
+    tree_name = tree['name']
+    tree_llvm_ver = tree['llvm_version']
+
+    # Calculate the number of times that a workflow runs in a week based on its
+    # schedule
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    week_from_now = now + datetime.timedelta(weeks=1)
```
nickdesaulniers, line 26: Loop invariant. Hoist this out of the `for` loop.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,57 @@
...
+sys.path.insert(0, str(ci_root))
+# pylint: disable-next=wrong-import-position
+from utils import get_config_from_generator  # noqa: E402
+
+config = get_config_from_generator()
+
+builds_per_tree = {}
+for tree in config['tree_schedules']:
+    tree_name = tree['name']
+    tree_llvm_ver = tree['llvm_version']
```
nickdesaulniers, line 21: Sink these closer to their first use, or get rid of them; this loop uses a mix of local vars and dict access inconsistently.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,57 @@
...
+    # schedule
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+    week_from_now = now + datetime.timedelta(weeks=1)
+    num_runs = len(
+        list(croniter.croniter_range(now, week_from_now, tree['schedule'])))
+
+    if tree_name not in builds_per_tree:
+        builds_per_tree[tree_name] = {'total': 0}
+    if tree_llvm_ver not in builds_per_tree[tree_name]:
+        builds_per_tree[tree_name][tree_llvm_ver] = 0
```
nickdesaulniers, line 33: I wonder if a `defaultdict` is more concise for these?
https://stackoverflow.com/a/9139961
It's nice when you have default values when a key does not exist.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,57 @@
...
+
+print(
+    f"Total builds per week: {sum(item['total'] for item in builds_per_tree.values())}",
+)
+
+for item in sorted(builds_per_tree.items(),
+                   key=lambda x: x[1]['total'],
+                   reverse=True):
+    tree = item[0]
+    builds = item[1]
```
nickdesaulniers, line 50: This is a bit convoluted.

I wonder if we can use destructed assignment rather than define `item`? Can we avoid the `key` parameter?

scripts/estimate-builds.py
```diff
@@ -0,0 +1,57 @@
...
+
+    for build in config['builds']:
+        if tree['git_repo'] == build['git_repo'] and \
+           tree['git_ref'] == build['git_ref'] and \
+           tree_llvm_ver == build['llvm_version']:
+            builds_per_tree[tree_name]['total'] += num_runs
+            builds_per_tree[tree_name][tree_llvm_ver] += num_runs
+
+print(
+    f"Total builds per week: {sum(item['total'] for item in builds_per_tree.values())}",
```
nickdesaulniers, line 43: Can we disable the trailing comma lint here as well?

scripts/estimate-builds.py
nathanchance: Done:

scripts/estimate-builds.py
nickdesaulniers: > Can we avoid the key parameter?

Ah, this is sorting by total builds descending. Consider adding a comment.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,58 @@
...
+                   reverse=True):
+    tree = item[0]
+    builds = item[1]
+
+    print(f"\n  - tree: {tree}")
+    print(f"    total: {builds['total']}")
+    print('    breakdown:')
+    del builds['total']
+    for k, v in builds.items():
+        print(f"    - clang-{k}: {v}")
```
nickdesaulniers, line 58: about more descriptive names like `clang_version` and `num_builds` rather than `k`, `v`?

scripts/estimate-builds.py
```diff
@@ -0,0 +1,58 @@
...
+for item in sorted(builds_per_tree.items(),
+                   key=lambda x: x[1]['total'],
+                   reverse=True):
+    tree = item[0]
+    builds = item[1]
+
+    print(f"\n  - tree: {tree}")
+    print(f"    total: {builds['total']}")
+    print('    breakdown:')
+    del builds['total']
```
nickdesaulniers, line 56: or only print if `k` is not `total`

scripts/estimate-builds.py
nickdesaulniers:
```diff
diff --git a/scripts/estimate-builds.py b/scripts/estimate-builds.py
index 973e92389c9a..a075022d7ed6 100755
--- a/scripts/estimate-builds.py
+++ b/scripts/estimate-builds.py
@@ -43,12 +43,9 @@ print(
     f"Total builds per week: {sum(item['total'] for item in builds_per_tree.values())}",
 )
 
-for item in sorted(builds_per_tree.items(),
+for tree, builds in sorted(builds_per_tree.items(),
                    key=lambda x: x[1]['total'],
                    reverse=True):
-    tree = item[0]
-    builds = item[1]
-
     print(f"\n  - tree: {tree}")
     print(f"    total: {builds['total']}")
     print('    breakdown:')
```

scripts/estimate-builds.py
```diff
@@ -0,0 +1,58 @@
...
+
+builds_per_tree = {}
+for tree in config['tree_schedules']:
+    tree_name = tree['name']
+    tree_llvm_ver = tree['llvm_version']
+
+    # Calculate the number of times that a workflow runs in a week based on its
+    # schedule
+    num_runs = len(
+        list(croniter.croniter_range(now, week_from_now, tree['schedule'])))
```
nickdesaulniers, line 29: Though we have many different builds, we only have a small number of schedules. Recomputing this number for each build is expensive. Consider throwing the result in a dict keyed on `tree['schedule']`.

scripts/estimate-builds.py
```diff
@@ -0,0 +1,58 @@
...
+
+import datetime
+from pathlib import Path
+import sys
+
+import croniter
+
+# Add the root of the repo to PYTHONPATH for utils
+ci_root = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ci_root))
```
nickdesaulniers, line 12: Should this go into `get_config_from_generator`? I assume this would be nice to have for any other caller?

scripts/estimate-builds.py
nathanchance: Sorry, copy and paste from the old version that had that linting applied: c3234a6841f9c339387a511cee69400606ab24b3

scripts/estimate-builds.py
nathanchance: Sure: e89a021ac9d4b4f01ed7f747f66c43109606db15

I opted for variables when I need the value from the dictionary in more than one place.

scripts/estimate-builds.py
nathanchance: Yeah, I think that looks nicer: e500f351ecfae53c84cf87f6cae10388ef903f13

scripts/estimate-builds.py
nathanchance: Thanks, I swore I tried unpacking this and failed but I was probably just doing something wrong:

d409f9327545ccc13408cd59f8794d7e84758549
c26df715981ac8ae23db4075d11d50d9d8316de6

scripts/estimate-builds.py
nathanchance: Sure, no strong opinion: 86254acd84d9066561d486fe9de4d78a5c5d44f4

scripts/estimate-builds.py
nathanchance: Sure thing: 7cb9ead20e11e32e5a5b092fce6fe9e478d88a38

scripts/estimate-builds.py
nickdesaulniers: oh! Indeed not worth it!

scripts/estimate-builds.py
```diff
@@ -0,0 +1,52 @@
...
+for tree, builds in sorted(builds_per_tree.items(),
+                           key=lambda x: x[1]['total'],
+                           reverse=True):
+    print(f"\n  - tree: {tree}")
+    print(f"    total: {builds['total']}")
+    print('    breakdown:')
+    for clang_version, num_builds in builds.items():
+        if clang_version == 'total':
+            continue
+        print(f"    - clang-{clang_version}: {num_builds}")
```
nickdesaulniers, line 52:
```python
if clang_version != 'total':
  print(...)
```

scripts/estimate-builds.py
```diff
@@ -0,0 +1,52 @@
...
+    for build in config['builds']:
+        if tree['git_repo'] == build['git_repo'] and \
+           tree['git_ref'] == build['git_ref'] and \
+           tree_llvm_ver == build['llvm_version']:
+            builds_per_tree[tree_name]['total'] += num_runs
+            builds_per_tree[tree_name][tree_llvm_ver] += num_runs
+
+print(
+    f"Total builds per week: {sum(item['total'] for item in builds_per_tree.values())}"
+)
```
nickdesaulniers, line 40: does this no longer fit on one line?

nathanchance: No, `yapf` will force this because the line is over 80 characters. I could do this:

```diff
diff --git a/scripts/estimate-builds.py b/scripts/estimate-builds.py
index a545768..3ac188a 100755
--- a/scripts/estimate-builds.py
+++ b/scripts/estimate-builds.py
@@ -35,9 +35,8 @@ for tree in config['tree_schedules']:
             builds_per_tree[tree_name]['total'] += num_runs
             builds_per_tree[tree_name][tree_llvm_ver] += num_runs

-print(
-    f"Total builds per week: {sum(item['total'] for item in builds_per_tree.values())}"
-)
+total_builds = sum(item['total'] for item in builds_per_tree.values())
+print(f"Total builds per week: {total_builds}")

 # Sort the list of builds by total number of builds descending
 for tree, builds in sorted(builds_per_tree.items(),
```

or perhaps we bump the line length to 100 like the kernel, it is not like we do not have the screen real estate nowadays but that is obviously a larger change for the whole repo. Up to you.

scripts/estimate-builds.py
nathanchance: Thanks, fixed: 4334205a69a025523c713beb0c4f192d33a39f4a

scripts/estimate-builds.py
nickdesaulniers: I prefer that diff, please apply it.

nickdesaulniers (approved):
thanks for the nice script!

scripts/estimate-builds.py
nickdesaulniers: From profiling `python3 -m cProfile -s cumtime ./scripts/estimate-builds.py | less` it looks like runtime is dominated by parsing the generator `utils.py:9(get_config_from_generator)`.

scripts/estimate-builds.py
nathanchance: Done: 4dde30ae0a0be9bf5671d7fc6106c882fd610cff

nathanchance:
Thank you for the expert review as always! All the linters are happy locally so I am merging this without presubmit finishing.

PR #530, opened 2023-03-08
Update patches (March 8, 2023)

nathanchance:
The android-mainline patches are now merged via an upstream update.

The arm64-fixes tree updated to 6.3-rc1 and picked up a mainline regression. I pinged the upstream thread again to try and get that change moving.


nickdesaulniers (approved):
LGTM; thanks! :moon: :star2:

PR #531, opened 2023-03-10
Copy stable patches to 6.1

nathanchance:
This change is needed in 6.1 as well but this was missed during the 6.1/6.2 split. It is currently pending in AUTOSEL so it should make its way to these trees shortly:

https://lore.kernel.org/all/20230303214315.1447666-57-sashal@kernel.org/
https://lore.kernel.org/all/20230303214106.1446460-61-sashal@kernel.org/

Additionally, this patch is now in mainline, so make that clear by renaming the patch to the upstream SHA and including a link to the commit on git.kernel.org.


nathanchance:
I guess these are out for review now:

https://lore.kernel.org/stable/20230310133722.384562012@linuxfoundation.org/
https://lore.kernel.org/stable/20230310133724.461889354@linuxfoundation.org/

I am not sure this workflow will run again before the next update so maybe we want to hold off on this? At the same time, I will have to drop the patches in 6.2 anyways so it probably makes sense to keep everything in sync.

nathanchance:
The patch has been merged and released, so I will push a new PR removing the patch from 6.1, rather than adding it to 6.2

PR #534, opened 2023-03-20
Drop mainline patches (March 20, 2023)

nathanchance:
Linus applied the patch directly:

https://git.kernel.org/linus/7d31677bb7b1944ac89e9155110dc1b9acbb3895


nickdesaulniers (approved):
go Linus, thanks for the anti-patch! :penguin: :wastebasket: :fire:

PR #536, opened 2023-03-22
Add powerpc allmodconfig build on linux-next

nathanchance:
In order to use ld.lld (and by extension, LLVM=1) with allmodconfig
(which is big endian), we need to build with the ELFv2 ABI, as ld.lld
does not have support for the ELFv1 ABI. This is now possible on -next,
so wire up this build to make sure it stays green.

This adds 15 more builds per week:

```
diff --git a/tmp/estimate-builds.txt b/tmp/.psub.GnbBewMeWv
index 24a4a39..ad40946 100644
--- a/tmp/estimate-builds.txt
+++ b/tmp/.psub.GnbBewMeWv
@@ -1,4 +1,4 @@
-Total builds per week: 8148
+Total builds per week: 8163

   - tree: mainline
     total: 2900
@@ -12,11 +12,11 @@ Total builds per week: 8148
     - clang-11: 520

   - tree: next
-    total: 2150
+    total: 2165
     breakdown:
-    - clang-17: 320
-    - clang-16: 320
-    - clang-15: 300
+    - clang-17: 325
+    - clang-16: 325
+    - clang-15: 305
     - clang-14: 285
     - clang-13: 285
     - clang-12: 265
```

### DIFF
--- a/.github/workflows/next-clang-15.yml
+++ b/.github/workflows/next-clang-15.yml
@@ -1243,6 +1243,27 @@ jobs:
         name: output_artifact_allconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
+  _516fac05d5416a9f78740a10f02ebe3f:
+    runs-on: ubuntu-latest
+    needs: kick_tuxsuite_allconfigs
+    name: ARCH=powerpc BOOT=0 LLVM=1 LLVM_IAS=1 LLVM_VERSION=15 allmodconfig+CONFIG_PPC64_BIG_ENDIAN_ELF_ABI_V2=y+CONFIG_WERROR=n
+    env:
+      ARCH: powerpc
+      LLVM_VERSION: 15
+      BOOT: 0
+      CONFIG: allmodconfig+CONFIG_PPC64_BIG_ENDIAN_ELF_ABI_V2=y+CONFIG_WERROR=n
+    container:
+      image: ghcr.io/clangbuiltlinux/qemu
+      options: --ipc=host
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+    - uses: actions/download-artifact@v3
+      with:
+        name: output_artifact_allconfigs
+    - name: Check Build and Boot Logs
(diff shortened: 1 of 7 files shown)


nickdesaulniers (commented):
are we able to build this target with clang-14 and older, too?

nathanchance:
I added a dependency on ld.lld 15.x+ to limit the chances of hitting old bugs in LLVM but I do not have any evidence that there are any. I can test with the older versions and change it upstream but until then, this configuration cannot be selected.

nickdesaulniers (approved):
Please file a todo to test that combination.

LGTM; thanks! :om: :woman_cartwheeling:

nathanchance:
This set of configurations works for LLVM 14.0.6 it seems but earlier versions (which do not have https://github.com/llvm/llvm-project/commit/2fb9029f2678e3698de202fa88150b3b03cf76da) error out when using the integrated assembler.

```
../arch/powerpc/kvm/book3s_hv_p9_entry.c:315:15: error: invalid instruction, did you mean: lwsync, sync?
        asm volatile("hwsync" ::: "memory");
                     ^
<inline asm>:1:2: note: instantiated into assembly here
        hwsync
        ^~~~~~
../arch/powerpc/kvm/book3s_hv_p9_entry.c:340:15: error: invalid instruction, did you mean: lwsync, sync?
        asm volatile("hwsync" ::: "memory");
                     ^
<inline asm>:1:2: note: instantiated into assembly here
        hwsync
        ^~~~~~
../arch/powerpc/kvm/book3s_hv_p9_entry.c:364:15: error: invalid instruction, did you mean: lwsync, sync?
        asm volatile("hwsync" ::: "memory");
                     ^
<inline asm>:1:2: note: instantiated into assembly here
        hwsync
        ^~~~~~
3 errors generated.
```

We could use GNU as for older versions if we wanted to gain that coverage.

Since we are only gaining one extra version of coverage,  I am tempted to just leave the upstream versions alone to avoid patch churn, but I do not have a strong opinion.

PR #537, opened 2023-03-27
Drop majority of Android patches (March 27, 2023)

nathanchance:
These patches are now in these Android branches via upstream LTS merges.

android12-5.4 can be dropped once
https://android-review.googlesource.com/c/kernel/common/+/2505475 is
merged.


nickdesaulniers (approved):
LGTM; thanks! :orange_circle: :red_circle:

Issue #540, opened 2023-03-28
build.log not fetched/displayed


nickdesaulniers:
https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/4527641306/jobs/7974728249#step:5:145  failed but did not print build.log.

https://storage.tuxsuite.com/public/clangbuiltlinux/continuous-integration2/builds/2NZv76Eis4NY8IEdQXmbhTNB59c/ should be the artifact for that build. Looks like it wasn't even fetched?

nathanchance:
We only print `build.log` if `errors_count` or `warnings_count` is greater than 0, which you can see from the JSON print out is not true. We can just adjust that condition to include if the build failed.

nickdesaulniers:
I'm inclined to treat this as a tuxsuite issue then; if the linker prints an error, then tuxsuite needs to set that in the metadata.json.

nickdesaulniers:
https://gitlab.com/Linaro/tuxsuite/-/issues/188

nathanchance:
The issue has been resolved upstream.

PR #542, opened 2023-03-28
check_logs.py: Print logs if build result is not a pass

nathanchance:
There are times where there could be build errors not caught by
tuxsuite's regex (like linker errors without "Error" in the string).

Print build.log if the result of the build is not a "pass".

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/540


nickdesaulniers:
Let's wait and see what the response to https://gitlab.com/Linaro/tuxsuite/-/issues/188 is first.

Issue #539, opened 2023-03-28
x86_64 allyesconfig builds of stable timing out


nickdesaulniers:
https://github.com/clangbuiltlinux/continuous-integration2/actions/workflows/stable-clang-17.yml

for 3 weeks now this has been timing out

nathanchance:
This is due to https://github.com/ClangBuiltLinux/linux/issues/1768 (I note in that issue why this is specific to `clang-15`+). We need Peter's series to be backported to 6.2 in order to fix this (or we can just disable `allyesconfig` for stable until it is updated to 6.3 in a few weeks).

nickdesaulniers:
Let's disable it for now then.

PR #543, opened 2023-03-28
Disable x86_64 allyesconfig builds on stable

nathanchance:
The build time performance regression from call depth tracking is not
fixed in this branch, causing the builds to timeout. Until that series
can be backported, disable these builds to keep the CI green.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/539

### DIFF
--- a/.github/workflows/stable-clang-15.yml
+++ b/.github/workflows/stable-clang-15.yml
@@ -1306,25 +1306,4 @@ jobs:
         name: output_artifact_allconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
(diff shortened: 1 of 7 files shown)


nickdesaulniers (approved):
thanks; let's leave an action item open to revert this once stable moves to 6.3. 🐳 🐋

nathanchance:
> thanks; let's leave an action item open to revert this once stable moves to 6.3. 🐳 🐋

Done, thanks!

https://github.com/ClangBuiltLinux/continuous-integration2/issues/544

PR #548, opened 2023-04-07
drop triples

nickdesaulniers:
https://lore.kernel.org/llvm/20230401170117.1580840-1-masahiroy@kernel.org/ reminded me that we probably don't need to set CROSS_COMPILE any more. Nathan confirmed that tuxsuite doesn't allow setting this, and that these triples aren't consumed or emitted by generate.py.


nathanchance (approved):
Good bye cruft!

PR #551, opened 2023-04-14
Update 64-bit big endian PowerPC configuration target

nathanchance:
A patch in review will change pseries_defconfig from a big endian
configuration to a little endian configuration:

https://lore.kernel.org/20230414132415.821564-32-mpe@ellerman.id.au/

As suggested in the commit message, switch to ppc64_guest_defconfig now
to guarantee that we always get a 64-bit big endian kernel that can boot
in QEMU, which is exactly what this configuration intends to do,
regardless of whether or not that change is actually accepted.

This configuration target is available on 5.4, the oldest kernel we
build ppc64 on, and it works fine with both clang-12 and clang-17, the
oldest and newest toolchain versions that this configuration is built
with.

### DIFF
--- a/.github/workflows/5.10-clang-12.yml
+++ b/.github/workflows/5.10-clang-12.yml
@@ -222,15 +222,15 @@ jobs:
         name: output_artifact_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _38b22cfae74996dcecea1459e2f4a5d1:
+  _b93c483101f871a5ff674333fdf22ffd:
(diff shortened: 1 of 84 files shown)


nickdesaulniers (approved):
LGTM; thanks! :robot: :moon:

mpe:
Yeah I think this would be a good change. Using `ppc64_guest_defconfig` will build more BE platforms and so give you more test coverage. A quick test here shows it taking ~5% longer to build, which seems acceptable.

PR #564, opened 2023-05-15
check_logs: fix RUF010 warnings about f-strings

nickdesaulniers:
Fixes:
```
check_logs.py:28:54: RUF010 [*] Use conversion in f-string
check_logs.py:37:54: RUF010 [*] Use conversion in f-string
```
Via:
`$ ruff check . --fix`


nickdesaulniers:
Perhaps related to https://github.com/charliermarsh/ruff/pull/4423. IDK what `!s` is in python...

Issue #574, opened 2023-05-25
856dd6c59812 for android-mainline


nickdesaulniers:
thumb2 builds of android-mainline are failing
https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/5076729153/jobs/9120297489#step:5:1262
they need
commit 856dd6c59812 ("ext4: fix unused iterator variable warnings")
if android-mainline isn't updated soon.

nathanchance:
I will give it a week to see if they can get close enough to 6.4-rc1 to include that change.

PR #575, opened 2023-05-30
Android mainline

nickdesaulniers:


nathanchance (approved):
I marked this as fixing #574.

PR #576, opened 2023-05-30
Array bounds

nickdesaulniers:


nathanchance:
We only need the final patch in that series to address the warnings from that issue, can we just apply that one? If you sucked these down with `b4`, pass `-P _` to your original command with `7d0571ef5dc87904008c325a942cfed24dbbf42e.1645513670.git.gustavoars@kernel.org` as the message ID argument.

GustavoARSilva:
> We only need the final patch in that series to address the warnings from that issue, can we just apply that one? If you sucked these down with `b4`, pass `-P _` to your original command with `7d0571ef5dc87904008c325a942cfed24dbbf42e.1645513670.git.gustavoars@kernel.org` as the message ID argument.

I agree with this.

The patch series seems to be correct. I just need to rework it to prove the maintainers that the before/after binary output does not disrupt the original implementation. So the final series might grow in number of patches, but the changes should remain the same.

nathanchance (approved):
LGTM!

PR #577, opened 2023-06-02
Update for moving rootfs images to GitHub releases in boot-utils

nathanchance:
This pull request is basically the second part of https://github.com/ClangBuiltLinux/boot-utils/pull/105.

The first change is just a clean up/update to consistently use absolute paths and pathlib, which is a bit more modern than the os methods we currently use and helps simplify things a bit in various places. This helps ensure everything is put in the correct place.

The second change downloads boot-utils straight from GitHub, which will allow us to drop the submodule check in, which will ultimately speed up the clones. After the aforementioned boot-utils pull request, the `*.py` files just need to reside in the same folder together, as they will automatically download the rootfs from GitHub.

The third change helps us avoid a theoretical rate limit from querying GitHub's API for the latest boot-utils images release. Rather than querying the API dynamically in every job that will be using boot-utils, we query it in the TuxSuite job, cache it via upload artifact, and pull it down in the child job, so the API will only be queried three times per workflow.

Finally, the final two changes remove the submodule and updates the workflows with the new steps.

See the commit messages for full details.

### DIFF
--- a/check_logs.py
+++ b/check_logs.py
@@ -112,8 +112,9 @@ def fetch_dtb(build):
         "multi_v5_defconfig": "aspeed-bmc-opp-palmetto.dtb",
         "aspeed_g5_defconfig": "aspeed-bmc-opp-romulus.dtb",
     }[config]
-    (dtb_path := Path(CI_ROOT, 'dtbs', dtb)).parent.mkdir(exist_ok=True)
-    url = build["download_url"] + dtb_path.name
+    remote_dtb_path = f"dtbs/{dtb}"
+    (dtb_path := Path(CI_ROOT, remote_dtb_path)).parent.mkdir(exist_ok=True)
+    url = build["download_url"] + remote_dtb_path
     _fetch("DTB", url, dtb_path)
 
 
--- a/check_logs.py
+++ b/check_logs.py
@@ -196,6 +196,12 @@ def print_clang_info(build):
     subprocess.run(parse_cmd, check=True)
 
 
+def fetch_boot_utils_file(file_path):
+    url = f"https://github.com/ClangBuiltLinux/boot-utils/raw/main/{file_path.name}"
+    _fetch(file_path.name, url, file_path)
+    file_path.chmod(0o755)
+
+
 def run_boot(build):
     cbl_arch = get_cbl_name()
     kernel_image = Path(CI_ROOT, get_image_name())
@@ -208,6 +214,10 @@ def run_boot(build):
         kernel_image.chmod(0o755)
     else:
         boot_cmd = [Path(boot_utils, 'boot-qemu.py'), "-a", cbl_arch]
+
+    fetch_boot_utils_file(boot_cmd[0])
+    fetch_boot_utils_file(Path(boot_utils, 'utils.py'))
+
     boot_cmd += [
         '--gh-json-file',
         Path(CI_ROOT, 'boot-utils.json'),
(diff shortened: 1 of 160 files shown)


check_logs.py
```diff
@@ -235,6 +230,11 @@ def run_boot(build):
                 "Disabling Oops problem matcher under Sanitizer KUnit build")
             print("::remove-matcher owner=linux-kernel-oopses::")
 
+    for dest in [boot_cmd[0], Path(boot_utils, 'utils.py')]:
+        url = f"https://github.com/ClangBuiltLinux/boot-utils/raw/main/{dest.name}"
+        _fetch(dest.name, url, dest)
+        dest.chmod(0o755)
```
nickdesaulniers, line 236: let's put this in a function and call it three times explicitly. Once for boot-uml.py or boot-qemu.py, then unconditionally for utils.py.  That makes it more obvious we are doing at most 2 fetches.

This logic should remain closer to the Path construction for these above (~L206). `boot_cmd[0]` and the distance between this logic and where `boot_cmd` is defined makes me nervous about an array push_front winding up inbetween in the future.

nickdesaulniers (approved):
Nice work!  Consider landing when you have time to manually kick a workflow, and fixup/revert should anything unforseen go immediately wrong.  Thanks again. 🌐 🚥

check_logs.py
nathanchance: > let's put this in a function and call it three times explicitly. Once for boot-uml.py or boot-qemu.py, then unconditionally for utils.py. That makes it more obvious we are doing at most 2 fetches.
> 
> This logic should remain closer to the Path construction for these above (~L206). `boot_cmd[0]` and the distance between this logic and where `boot_cmd` is defined makes me nervous about an array push_front winding up inbetween in the future.

Would this diff address your concerns?

```diff
diff --git a/check_logs.py b/check_logs.py
index e20121c..9dda8bc 100755
--- a/check_logs.py
+++ b/check_logs.py
@@ -196,6 +196,12 @@ def print_clang_info(build):
     subprocess.run(parse_cmd, check=True)


+def fetch_boot_utils_file(file_path):
+    url = f"https://github.com/ClangBuiltLinux/boot-utils/raw/main/{file_path.name}"
+    _fetch(file_path.name, url, file_path)
+    file_path.chmod(0o755)
+
+
 def run_boot(build):
     cbl_arch = get_cbl_name()
     kernel_image = Path(CI_ROOT, get_image_name())
@@ -208,6 +214,10 @@ def run_boot(build):
         kernel_image.chmod(0o755)
     else:
         boot_cmd = [Path(boot_utils, 'boot-qemu.py'), "-a", cbl_arch]
+
+    fetch_boot_utils_file(boot_cmd[0])
+    fetch_boot_utils_file(Path(boot_utils, 'utils.py'))
+
     boot_cmd += [
         '--gh-json-file',
         Path(CI_ROOT, 'boot-utils.json'),
@@ -230,11 +240,6 @@ def run_boot(build):
                 "Disabling Oops problem matcher under Sanitizer KUnit build")
             print("::remove-matcher owner=linux-kernel-oopses::")

-    for dest in [boot_cmd[0], Path(boot_utils, 'utils.py')]:
-        url = f"https://github.com/ClangBuiltLinux/boot-utils/raw/main/{dest.name}"
-        _fetch(dest.name, url, dest)
-        dest.chmod(0o755)
-
     # Before spawning a process with potentially different IO buffering,
     # flush the existing buffers so output is ordered correctly.
     sys.stdout.flush()
```

nathanchance: 66e4a285290e2f2f620141f564b2059dff7504d2

nathanchance:
I have kicked off a presubmit run, might take some time for it to finish, I will address any issues that it finds then merge this.

https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/5225765696

nickdesaulniers:
Doesn't this need to be merged first before the run, to see if this PR breaks anything?

nathanchance:
I pushed my branch to this repo as `presubmit/577` and started a workflow manually from there so that there should be no disruptions to main while still getting valid results.

nathanchance:
85c2476dda8b24e8c058bc3fc6be7a3ab391a635 deals with the only issue presubmit sees, so I am merging this.

PR #580, opened 2023-06-08
Update Arch Linux x86_64 configuration link

nathanchance:
Arch Linux moved their packaging from svn to git, so the
svntogit-packages repo on GitHub is not being updated anymore. Switch
the link to the configuration in the git repo, so that we continue to
get configuration updates.


nickdesaulniers (approved):
LGTM; thanks! 🧇 🌔

PR #584, opened 2023-06-20
check_logs.py: Remove unnecessary conversion to fix updated RUF010

nathanchance:
15103a6 ("check_logs: fix RUF010 warnings about f-strings") added a conversion to silence a Ruff warning (`RUF010`). That warning has been updated in 0.0.273 to note that `f"{foo!s}"` is equivalent to `f"{foo}"`. The conversion is only needed if there is a format spec (such as a length).

This is the result of:

```
$ ruff --fix .
```


nickdesaulniers (approved):
ah thank god, that `!s` is fugly!

PR #587, opened 2023-06-27
patches: mainline: Add patch for RANDSTRUCT raid1-10.c error

nathanchance:
A patch broken under clang RANDSTRUCT went into mainline without the fix
for the error. It is now on the path to mainline but it may take some
time. To catch other potential breakage in the meantime, apply the
fix patch.


nickdesaulniers:
@knurd what's the best way to report these kinds of things?

nathanchance:
For what it's worth, this patch was already on its way to mainline when I applied it here, which is why I did not bother saying anything upstream:

https://lore.kernel.org/343874ee-9b2c-c333-e270-59c0c6a0244f@kernel.dk/

There was an additional report later, which did not cause the patch to be accelerated, which is pretty much what I expected to happen.

https://lore.kernel.org/ZJ2H5FWuo9oDMgPm@debian/

knurd:
> @knurd what's the best way to report these kinds of things?

Well, you could add them to the list of tracked regressions. But due to things like "priorities" and "there are only 24 hours in a day" I don't make much noise about build errors during the merge window, as everybody knows things are a bit chaotic. But that changes once we get closer to the end and in fact prodded one such issue today[1] someone added to the tracking in the hope to get it fixed before -rc1.

[1] https://lore.kernel.org/all/7493c624-431c-a67d-3cbd-10f29a8c33dd@leemhuis.info/

> There was an additional report later, which did not cause the patch to be accelerated,

Yup, but it seems a sub-maintainer did something stupid[2], otherwise all this likely would have worked way more smoothly.  

[2] https://lore.kernel.org/all/83240030-681c-9ff5-6e2c-600e83b0cc71@kernel.dk/

nickdesaulniers:
> Well, you could add them to the list of tracked regressions.

Sure, how do I do that, for next time?  Is that the regzbot comments?  [Gitlab is actively having a incident ATM](https://status.gitlab.com/), but I'll take a look at https://gitlab.com/knurd42/regzbot/-/blob/main/docs/getting_started.md when that's resolved.

> [1] https://lore.kernel.org/all/7493c624-431c-a67d-3cbd-10f29a8c33dd@leemhuis.info/

As a heads up, https://lore.kernel.org/all/13abe42f-2f5f-cbaf-21b8-baa4516963aa@leemhuis.info/ mentions that the report is a duplicate of https://lore.kernel.org/all/ZJ2M4yqnOCqqGWH0@debian/. They are **not** duplicates; those are distinct breakages.

> But due to things like "priorities" and "there are only 24 hours in a day" I don't make much noise about build errors during the merge window, as everybody knows things are a bit chaotic.

No worries; I figure it will get sorted out quickly. More so I'm interested in tracking incidents like this for posterity, since then at plumbers I can point out the frequency with which this kind of issue occurs.

knurd:
> > Well, you could add them to the list of tracked regressions.
>> Sure, how do I do that, for next time? [...]

Yeah, https://gitlab.com/knurd42/regzbot/-/blob/main/docs/getting_started.md explains the details. But if you just want just the important bits, look here:

https://docs.kernel.org/admin-guide/reporting-regressions.html
https://docs.kernel.org/process/handling-regressions.html

> As a heads up, […] They are **not** duplicates; those are distinct breakages.

Yeah, had noticed that stupid error a few minutes later and fixed it up manually. Thx for the heads up nevertheless.

> since then at plumbers I can point out the frequency with which this kind of issue occurs.

FWIW, I have no numbers, but I got the impression that we had quite a bit more build errors this merge window. Wonder how much trouble that will cause for people performing a bisection – and if things like the flexible-array member stuff should be merged differently (at the end of the merge window with all fixes that subsystems maintainers did not pick up maybe?) to reduce them.

nathanchance:
> FWIW, I have no numbers, but I got the impression that we had quite a bit more build errors this merge window. Wonder how much trouble that will cause for people performing a bisection – and if things like the flexible-array member stuff should be merged differently (at the end of the merge window with all fixes that subsystems maintainers did not pick up maybe?) to reduce them.

I would estimate that there is some sort of build error that breaks our CI close to every -next release during a development cycle; for example, our `clang-16` builds were red most of the 6.5 development cycle:

https://github.com/clangbuiltlinux/continuous-integration2/actions/workflows/next-clang-16.yml

It is not until the merge window that things finally tend to stabilize and even then, we are applying a patch for a build warning that gets promoted to an error with `CONFIG_WERROR=y` right now because the maintainers did not want to pick it up until after -rc1:

https://github.com/ClangBuiltLinux/continuous-integration2/tree/main/patches/next

https://lore.kernel.org/567481c3-08bc-3bb8-5538-0d296ee539d9@xs4all.nl/

That one is not even clang specific, it reproduces with GCC:

https://lore.kernel.org/202307070653.xxncAC2W-lkp@intel.com/

https://lore.kernel.org/bd856932-c2fb-d5f2-17f7-f7f1a1773adf@linux-m68k.org/

PR #588, opened 2023-06-28
patches: mainline: Add patch for drivers/scsi/aacraid

nathanchance:
Same as #576 but for mainline, as `-fstrict-flex-arrays=3` has been merged, causing `-Warray-bounds` in `drivers/scsi/aacraid/commsup.c`.

There is another set of warnings in `drivers/staging/ks7010` but the fix for that is pending in a pull request, so I am not adding it here with the assumption that it will be pulled shortly (the wireless patch):

https://lore.kernel.org/ZJxZJDUDs1ry84Rc@work/

cc @kees @GustavoARSilva


GustavoARSilva:
I have the rebase series ready for this here:

https://git.kernel.org/pub/scm/linux/kernel/git/gustavoars/linux.git/log/?h=testing/fam01-next20230627

I'll send it out this afternoon. :)

GustavoARSilva:
Here is the new series:

https://lore.kernel.org/linux-hardening/cover.1687974498.git.gustavoars@kernel.org/T/

nathanchance:
Thanks! I have updated the patch, which is just metadata but it will help with tracking.

PR #590, opened 2023-06-29
build-local.py improvements

nathanchance:
This pull request adds some quality of life improvements to `scripts/build-local.py`.

The first commit allows local toolchains to be used with `tuxmake` by having something like:

```yaml
toolchain: clang
make_variables:
  LLVM: 1
```

or

```yaml
toolchain: gcc
make_variables:
  CROSS_COMPILE: aarch64-linux-
```

then running `build-local.py` like

```
$ PATH=$TC_PREFIX_1/bin:$TC_PREFIX_2/bin:$PATH scripts/build-local.py ...
```

This is useful for using the versions of GCC and LLVM available on [kernel.org](https://kernel.org/pub/tools) with `tuxmake`. `podman` or `docker` is not required in this instance, so the error is made conditional on whether or not versioned toolchains (ones with a `-` in them) are actually requested.

The second patch allows interrupting `build-local.py` with Ctrl-C. The current behavior just kills `tuxmake` internally and the iteration of builds continue, rather than being immediately stopped.

The third patch allows the use of `ccache` if it is available in the user's environment.

### DIFF
--- a/scripts/build-local.py
+++ b/scripts/build-local.py
@@ -73,9 +73,7 @@ if shutil.which('podman'):
 elif shutil.which('docker'):
     runtime = 'docker'
 else:
-    raise RuntimeError(
-        "tuxmake requires either podman or docker for versioned toolchains but neither could be found on your system!"
-    )
+    runtime = None
 
 # Combine all jobs into one object for easy iteration
 jobs = {}
@@ -109,6 +107,11 @@ for name, builds in jobs.items():
         kconfig = build['kconfig']
         target_arch = build['target_arch']
         toolchain = build['toolchain']
+
+        if '-' in toolchain and not runtime:
+            raise RuntimeError(
+                f"tuxmake requires either podman or docker to use versioned toolchains ('{toolchain}') but neither could be found on your system!"
+            )
 
         cfg_str = '+'.join(kconfig) if isinstance(kconfig, list) else kconfig
         print(f"I: Building {target_arch} {cfg_str} ({toolchain})... ",
--- a/scripts/build-local.py
+++ b/scripts/build-local.py
@@ -5,6 +5,7 @@ from argparse import ArgumentParser
 import copy
 from pathlib import Path
 import re
+import signal
 import shutil
 import sys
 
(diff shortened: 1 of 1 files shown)


scripts/build-local.py
```diff
@@ -110,6 +108,11 @@
         target_arch = build['target_arch']
         toolchain = build['toolchain']
 
+        if '-' in toolchain and not runtime:
```
nickdesaulniers, line 136: where does the `'-'` come from?

nathanchance: I made a small comment about this above:

> so the error is made conditional on whether or not versioned toolchains (ones with a `-` in them) are actually requested.

Should I clarify with a code comment too?

scripts/build-local.py
```diff
@@ -14,6 +15,16 @@
 RED = '\033[01;31m'
 GREEN = '\033[01;32m'
 
+
+def interrupt_handler(_signum, _frame):
+    """
+    Allows script to exit immediately when Ctrl-C is pressed
+    """
+    sys.exit(130)
```
nickdesaulniers, line 24: FWIW, `os` contains more descriptive error codes
https://docs.python.org/3/library/os.html#os._exit
not sure it's required here though.

nickdesaulniers (approved):

scripts/build-local.py
nickdesaulniers: ah, in the PR description. Ok let me reparse what you said.

nickdesaulniers: yes, I think a comment in the code here would help clarify or at least help me remember what's going on here in the future.

scripts/build-local.py
nathanchance: Perhaps

```diff
diff --git a/scripts/build-local.py b/scripts/build-local.py
index 23c760c..7c33198 100755
--- a/scripts/build-local.py
+++ b/scripts/build-local.py
@@ -21,7 +21,7 @@ def interrupt_handler(_signum, _frame):
     """
     Allows script to exit immediately when Ctrl-C is pressed
     """
-    sys.exit(130)
+    sys.exit(signal.SIGINT)


 signal.signal(signal.SIGINT, interrupt_handler)
```

I chose 130 because that is what shells have the return code to when Ctrl-C is pressed for me in the past.

PR #591, opened 2023-06-30
Apply series to fix arm64 big endian on stable

nathanchance:
Same as commit 872fc01 but for stable, which has the same issue.


nickdesaulniers:
These are now in mainline; time to send a request to stable?

nathanchance:
> These are now in mainline; time to send a request to stable?

I assume you'll see it soon ;) :P

https://lore.kernel.org/20230703150339.GA1975402@dev-arch.thelio-3990X/

nickdesaulniers:
heh, yep, going through my inbox out of order. Thanks!

PR #593, opened 2023-06-30
Move powerpc64 to LLVM=1 on mainline/-next with LLVM 14.x+

nathanchance:
After commit [9d90161ca5c7](https://git.kernel.org/linus/9d90161ca5c7234e80e14e563d198f322ca0c1d0) ("powerpc/64: Force ELFv2 when building with LLVM linker"), ELFv2 is the default when linking with `ld.lld`, which allows us to switch to normal `LLVM=1` when building `ppc64_guest_defconfig` on mainline and -next. IAS requires https://github.com/llvm/llvm-project/commit/33504b3bbe10d5d4caae13efcb99bd159c126070 from LLVM 14.x, so `LLVM_IAS=0` is still necessary for those builds.

I have qualified this locally with `scripts/build-local.py`.

### DIFF
--- a/.github/workflows/mainline-clang-12.yml
+++ b/.github/workflows/mainline-clang-12.yml
@@ -449,10 +449,10 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
-  _b93c483101f871a5ff674333fdf22ffd:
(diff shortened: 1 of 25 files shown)


nickdesaulniers (approved):
nice, maybe worth updating Documentation/kbuild/llvm.rst in particular, changing powerpc from CC=clang to LLVM=1.

PR #601, opened 2023-07-12
Update SCSI -Warray-bounds patch application

nathanchance:
Update the patch to the final version in mainline and apply it to -tip. See commits for individual details.


nickdesaulniers (approved):
excellent, thank you! :mermaid: :merperson:

PR #602, opened 2023-07-13
Update arm64-fixes patches and update verisilicon patch

nathanchance:
Apply the mainline patches for the issues that impact ARCH=arm64 until
the arm64-fixes fast forwards to a revision that includes them (neither
are currently merged).

Update the verisilicon patch, which helps with tracking, as the patch is no
longer on the mailing list but in a maintainers tree, which should have a
stable SHA.


nickdesaulniers (approved):
LGTM; thanks! 🚋 🚂

PR #605, opened 2023-07-17
add raid1-10 patch for randstruct for android mainline

nickdesaulniers:


nathanchance (approved):
I was hoping to avoid patching this tree since 6.5-rc1 should eventually be merged up but that could take some time depending on the regressions or issues that they hit.

PR #609, opened 2023-07-26
Update LLVM tip of tree anchor to 18

nathanchance:
While clang-nightly in TuxSuite has not updated to clang-18 just yet, it
should be soon and this is needed so that updating the generated files
works properly.


nickdesaulniers (approved):
LGTM; thanks! :llama: :v:

PR #611, opened 2023-07-27
patches: arm64: Update for 6.5-rc3 fast forward

nathanchance:
The arm64 tree has now updated to 6.5-rc3, which allows us to drop the `--target=` to `KBUILD_CPPFLAGS` series that we were applying.

Unfortunately, we have to apply commit [7ee8acd1b803](https://git.kernel.org/linus/7ee8acd1b803502878992acd6f99e61f1e8c7a25) ("media: verisilicon: fix excessive stack usage") to avoid an instance of `-Wframe-larger-than`, which is fatal with `CONFIG_WERROR`.


nickdesaulniers (approved):
LGTM; thanks! :runner: :running_shirt_with_sash:

Issue #524, opened 2023-02-27
wire up arm64 virtconfig


nickdesaulniers:
this is a newer defconfig for arm64

nathanchance:
Done with https://github.com/ClangBuiltLinux/continuous-integration2/pull/613.

PR #613, opened 2023-08-06
Apply virtconfig to defconfig arm64

phongt:
Apply for #524 

Change generator.yml  and for mainline branch only because virtconfig is available in v6.3-rc1

### DIFF
--- a/.github/workflows/mainline-clang-16.yml
+++ b/.github/workflows/mainline-clang-16.yml
@@ -1507,6 +1507,30 @@ jobs:
         name: boot_utils_json_allconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
+  _222fc780c5ec26a6b47ceb85ad54844c:
+    runs-on: ubuntu-latest
+    needs: kick_tuxsuite_allconfigs
+    name: ARCH=arm64 LLVM=1 LLVM_IAS=1 LLVM_VERSION=16 virtconfig
+    env:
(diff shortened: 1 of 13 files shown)


nathanchance (changes requested):
Thanks for the patch!

However, I don't think this change is quite right. I think `virtconfig` should be its own build target, otherwise we are losing build coverage, as [`virt.config` turns off a bunch of configurations](https://elixir.bootlin.com/linux/v6.5-rc5/source/arch/arm64/configs/virt.config) that are not necessary for running in a virtual machine. Truthfully, given that `virtconfig` is [by definition a subset of `defconfig`](https://elixir.bootlin.com/linux/v6.5-rc5/source/arch/arm64/Makefile#L210), I am not really sure there is much value in testing both.

If we do want to test both, I think we should just copy the `arm64` anchor, name it `arm64_virt`, change the target to `virtconfig`, and add it to all trees that support it. The automation notes that this change as it stands applies to all uses of `arm64`, which is not correct for older trees.

phongt:
just change again with new `arm64_virt` and apply for mainline/next with llvm tot/llvm-latest/llvm-16

nathanchance (approved):
The comment about what this actual gains us remains but otherwise, seems good to me! Thanks!

nickdesaulniers:
sorry for not merging this sooner! I had forgotten about it!

PR #614, opened 2023-08-07
patches: Add fix for new -Wconstant-conversion in clang-18

nathanchance:
This patch is needed to avoid build failures in `all{mod,yes}config` in trees that have `CONFIG_WERROR`.


nickdesaulniers (approved):
LGTM; thanks! :billed_cap: :owl:

PR #619, opened 2023-08-25
build x86_64 allmodconfig linux-5.4.y with clang-13+

nickdesaulniers:


generator.yml
```diff
@@ -1217,6 +1218,7 @@ builds:
   - {<< : *ppc64,             << : *stable-5_4,       << : *ppc64_llvm,      boot: true,  << : *llvm_latest}
   - {<< : *ppc64le,           << : *stable-5_4,       << : *llvm,            boot: true,  << : *llvm_latest}
   - {<< : *x86_64,            << : *stable-5_4,       << : *llvm_full,       boot: true,  << : *llvm_latest}
+  - {<< : *x86_64_allmod,     << : *stable-5_4,      << : *llvm_full,       boot: false, << : *llvm_latest}
```
nickdesaulniers, line 1221: need to fix indentation

JustinStitt:
How do we properly test these workflows pre-merge? The YAML configs look fine but I am not super familiar with testing workflows on repos I don't own.

nickdesaulniers:
In general, we don't run the actual test since they are relatively expensive to run in pre-submit. Instead we back out changes if they fail in post submit.

It is possible for reviewers to run the specific builds of the specific branches of the specific trees with the specific toolchain versions, if they were inclined.  But that's a bit too much to ask, and I don't mind merging, then letting our CI tell us later if something needs to be backed out.

JustinStitt (approved):

Issue #599, opened 2023-07-10
Add `ARCH=loongarch`


nathanchance:
Requires TuxMake/TuxSuite support first: https://gitlab.com/Linaro/tuxmake/-/issues/205

Once it is supported there, the following configurations need to be disabled to build at the time of writing this:

- https://github.com/ClangBuiltLinux/linux/issues/1883
```
CONFIG_CRASH_DUMP=n
CONFIG_RELOCATABLE=n
```
- https://github.com/ClangBuiltLinux/linux/issues/1884
```
CONFIG_MODULES=n
```

Additionally, [the support for LoongArch in `ld.lld`](https://reviews.llvm.org/D138135) needs to be merged to use full `LLVM=1`, otherwise we may be able to use GNU ld but I have not tested that yet.

nathanchance:
> Requires TuxMake/TuxSuite support first: https://gitlab.com/Linaro/tuxmake/-/issues/205

https://gitlab.com/Linaro/tuxmake/-/merge_requests/325

That is only one piece, we need `ld.lld` support merged into LLVM main to allow us to use `clang-nightly`, as Debian does not ship `loongarch64-linux-gnu` binutils or GCC.

xen0n:
FYI: the LLD LoongArch port [has been upstreamed](https://github.com/llvm/llvm-project/commit/6084ee742064cf8121444febee342133b0cb7abd) just in time for the LLVM 17 release. We will continue the work on enabling the other bits ;-)

nathanchance:
I have opened https://gitlab.com/Linaro/tuxsuite/-/issues/194 as a prerequisite.

nickdesaulniers:
commit 5a31ed4678e0 ("LoongArch: Mark Clang LTO as working")

so we should add LTO builds, too

nickdesaulniers:
What's the latest on this?

nathanchance:
This is blocked on https://gitlab.com/Linaro/tuxmake/-/issues/207 now.

PR #621, opened 2023-08-28
Update stable anchor to 6.5

nathanchance:
This is the latest stable branch.

PR #622, opened 2023-08-28
Add support for LoongArch

nathanchance:
This is only enabled for clang-nightly at the moment but clang-17 should have it as well, which can be done after https://gitlab.com/Linaro/tuxmake/-/merge_requests/338 is merged and available in `tuxsuite`.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/599

### DIFF
--- a/.github/workflows/mainline-clang-18.yml
+++ b/.github/workflows/mainline-clang-18.yml
@@ -473,6 +473,54 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
+  _c14d023a315cad36e5d5d6ea80436cc5:
+    runs-on: ubuntu-latest
+    needs: kick_tuxsuite_defconfigs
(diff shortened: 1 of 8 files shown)


nathanchance:
Actually, the `allyesconfig` builds will be broken due to https://github.com/ClangBuiltLinux/linux/issues/1923 but I would like to avoid turning off `CONFIG_WERROR=n` if possible...

nickdesaulniers (approved):
woah, thinLTO builds and boots! Nice! If full lto doesn't, let's get some bugs on file?

Thanks for the patch! :loop: :bow_and_arrow:

nathanchance:
> If full lto doesn't, let's get some bugs on file?

I had not tested full LTO but sure enough :) https://github.com/ClangBuiltLinux/linux/issues/1924

I will merge #621 first then rebase this PR onto latest main so that I can add this same coverage to `stable` then merge.

PR #624, opened 2023-09-05
Update checkout action version from 3 to 4 and two jobs' Ubuntu version from 20.04 to latest

nathanchance:
The [checkout action](https://github.com/actions/checkout)'s major version has been bumped from 3 to 4. Update our files accordingly.

In that process, I noticed two jobs were using `ubuntu-20.04` instead of `ubuntu-latest`. Adjust those accordingly.


nickdesaulniers (approved):
LGTM; thanks! :bug: :lady_beetle:

PR #625, opened 2023-09-05
patches: arm64-fixes: Update for 6.5-rc3 fast forward

nathanchance:
This drops the scsi -Warray-bounds patch and updates the hash of the
media patch.


nickdesaulniers (approved):
LGTM; thanks! :mechanical_arm: :foot:

PR #627, opened 2023-09-07
Fix LoongArch LTO builds

nathanchance:
I missed that CONFIG_FTRACE needs to be disabled because LoongArch uses
recordmcount, which is not supported when building with LTO.


nickdesaulniers (approved):
LGTM; thanks! :eyes: :loop:

PR #630, opened 2023-09-18
Patch refresh

nickdesaulniers:


nickdesaulniers:
example of patch application failure: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/6226336737/job/16898907319#step:6:72

PR #631, opened 2023-09-20
Drop test_scanf change from all stable trees

nathanchance:
This change is in 6.5.4, 6.1.54, and 5.15.132.


nickdesaulniers (approved):
LGTM; thanks! :rabbit: :twisted_rightwards_arrows:

PR #632, opened 2023-09-22
drop lib: test_scanf patch

nickdesaulniers:


nickdesaulniers:
example failure to apply: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/6242435717/job/16946520267#step:6:72

nathanchance:
I think that failure should have been taken care of by #631? I think this is still needed for arm64, as `for-next/core` has not fast forwarded to a 6.6 tag.

nickdesaulniers:
I see, failures in runs since then are addressed by the message you just sent to stable. Thanks!

PR #636, opened 2023-09-29
Android srso

nickdesaulniers:
These issues seem dependent on LLD version.

The fixes are in v5.10.191 and v5.10.192. ACK just upgraded to
v5.10.189, and isn't planning another uprev soon.  Let's just carry
these build break fixes because I worry about backporting them without
the rest of the series.


nathanchance (approved):
I think this is fine.

PR #637, opened 2023-10-02
add 5.10 drm/mtk patch

nickdesaulniers:
Stable is a bit behind picking up
https://lore.kernel.org/llvm/20230922-5-10-fix-drm-mediatek-backport-v1-1-912d76cd4a96@kernel.org/


nathanchance:
This is now out for review so we can revert this in two days or so: https://lore.kernel.org/20231009130131.123904096@linuxfoundation.org/

PR #640, opened 2023-10-06
Drop 5.15 fix up patch for drivers/interconnect/core.c

nathanchance:
It has been released as part of 5.15.134.


nickdesaulniers (approved):
LGTM; thanks! :stadium: :kenya:

PR #644, opened 2023-10-13
Relax CONFIG_BPF_PRELOAD=n to older stable trees

kees:
With issue https://github.com/ClangBuiltLinux/linux/issues/1433 fixed upstream, we don't need to exclude CONFIG_BPF_PRELOAD from the Fedora config builds any more for modern kernels. Leave it in place for the older stable builds, though.

### DIFF
--- a/generator.yml
+++ b/generator.yml
@@ -318,8 +318,9 @@ configs:
   - &arm64_allno       {config: allnoconfig,                                                                                               ARCH: *arm64-arch,      << : *default}
   - &arm64_allyes      {config: allyesconfig,                                                                                              ARCH: *arm64-arch,      << : *default}
   - &arm64_alpine      {config: *arm64-alpine-config-url,                                                                                  ARCH: *arm64-arch,      << : *kernel}
-  # CONFIG_BPF_PRELOAD disabled for all cross compiled Fedora configs: https://github.com/ClangBuiltLinux/linux/issues/1433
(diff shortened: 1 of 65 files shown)


generator.yml
```diff
@@ -572,7 +573,7 @@ builds:
   - {<< : *arm64_allno,       << : *stable,           << : *llvm_full,       boot: false, << : *llvm_tot}
   - {<< : *arm64_allyes,      << : *stable,           << : *llvm_full,       boot: false, << : *llvm_tot}
   - {<< : *arm64_alpine,      << : *stable,           << : *llvm_full,       boot: true,  << : *llvm_tot}
-  - {<< : *arm64_fedora,      << : *stable,           << : *llvm_full,       boot: true,  << : *llvm_tot}
+  - {<< : *arm64_fedora_bpf,  << : *stable,           << : *llvm_full,       boot: true,  << : *llvm_tot}
```
nathanchance, line 576: Was this intentional? I think this was resolved with https://git.kernel.org/linus/e96f2d64c812d9c20adea38a9b5e08feaa21fcf5, which landed in 5.18, so I think this particular conversion can be removed because `stable` is just "the latest stable release".

nathanchance (approved):
LGTM with one comment below. I can handle tackling the other architectures that disable this on Monday.

generator.yml
kees: Oops, no, that was a read-o. And yes, I totally failed to adjust the other architectures! Let me fix this properly...

kees:
Okay, did the other archs and fixed the one I missed.

nathanchance (approved):

PR #645, opened 2023-10-26
Disable arm64 big endian builds with LLVM 13 and 14

nathanchance:
LLVM's integrated assembler incorrect generates the encoding for nop when targeting arm64 big endian. This was fixed in LLVM 15, so disable builds for LLVM 13 and 14.

Link: https://github.com/ClangBuiltLinux/linux/issues/1948

NOTE: Instead of outright disabling these builds, we could move them to `LLVM=1 LLVM_IAS=0` instead. I do not have a strong opinion either way.


nickdesaulniers (approved):
thanks for the patch! :tractor: :beach_umbrella:

PR #646, opened 2023-10-27
patches: Add backports for new ld.lld issue in Android

nathanchance:
These have been sent upstream but are unlikely to end up in Android
super quickly due to these branches being released branches.

Link: https://github.com/ClangBuiltLinux/linux/issues/1949


nickdesaulniers (approved):
LGTM; thanks!

PR #648, opened 2023-11-01
Add fix for -Wc23-extensions in net/ipv4/tcp_output.c

nathanchance:


nickdesaulniers (approved):
thanks for the quick fix! :hand_over_mouth: :field_hockey:

PR #649, opened 2023-11-02
Update mainline and stable patches (November 2, 2023)

nathanchance:
Just two patches for recent build failures from changes on both the Linux and LLVM side.

I have opted not to patch next to avoid losing coverage if patches no longer apply (as they will end up there first).


nathanchance:
Added two more patches that are needed after last night's run: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/6745136000

nickdesaulniers (approved):
fricken' merge window... :mermaid: :window:

PR #650, opened 2023-11-06
remove 1dc05a274a7b

nickdesaulniers:
it's now in mainline and failing to apply.


nathanchance (approved):
Yeah, #613 was created before #624, so the builds added by it did not have `checkout@v4`.

PR #652, opened 2023-11-06
Copy mainline patches to tip (Nov 6, 2023)

nathanchance:
-tip has updated to 4c975a43fa38, so copy all the mainline patches over, as they have the same issues now.


patches/tip/series
```diff
@@ -0,0 +1,4 @@
+20231031_nathan_tcp_fix_wc23_extensions_in_tcp_options_write.patch
+cba4590036855f4e3110d43c14385d2401080dbb.patch
+v2_20231102_nathan_drm_amd_display_increase_frame_warning_limit_with_kasan_or_kcsan_in_dml2.patch
+v2_20231102_nathan_loongarch_mark___percpu_functions_as_always_inline.patch
```
nickdesaulniers, line 4: we don't need this since we don't build ARCH=loongarch on tip

nathanchance: Good catch, fixed in the latest push!

nickdesaulniers (approved):
LGTM; thanks! :bullettrain_side: :railway_track:

PR #655, opened 2023-11-06
utils.py: Handle virtconfig

nathanchance:
While virtconfig is currently an aarch64-only target, it could in theory become a target for other architectures as well, so handle it like just regular defconfigs.


nickdesaulniers (approved):
LGTM; thanks! :mechanical_arm: :virgo:

PR #656, opened 2023-11-08
patches: mainline: Drop AMDGPU -Wframe-larger-than bump patch

nathanchance:
It is now in mainline:

https://git.kernel.org/linus/6740ec97bcdbe96ac7df147f986c030eddfebe65


nickdesaulniers (approved):
LGTM; thanks! 🎗️ 📦

JustinStitt:
How do we identify when these patches are in mainline via the CI? I understand the build will fail because it may not be able to apply but what sort of error message is produced in the CI job log?

PR #657, opened 2023-11-09
Add hardening.config

phongt:
Apply for branch: kernel mainline and next
config: [arm64 hardening.config](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/configs/hardening.config), [x86_64 hardening.config](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/x86/configs/hardening.config)

### DIFF
--- a/.github/workflows/mainline-clang-18.yml
+++ b/.github/workflows/mainline-clang-18.yml
@@ -425,6 +425,30 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: ./check_logs.py
+  _ec1d9cf8c283eb375da6d050fc464943:
+    runs-on: ubuntu-latest
+    needs: kick_tuxsuite_defconfigs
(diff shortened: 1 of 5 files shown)


kees:
Hm, I think allmodconfig is a superset of the hardening.config configs -- is there a useful difference here where we gain meaningful build coverage?

nathanchance:
I suppose boot testing would be the major gain because we don't boot test allmodconfig. However, it'd be interesting to see the difference of what actually gets enabled in defconfig and see if those configs are really worth testing in our simple VMs.

kees:
Yeah, boot testing does make sense here. I hadn't considered that. Yeah, let's do it! :)

nathanchance (approved):
LGTM, thanks! If this works out for LLVM ToT, I think we should turn this on for stable LLVM as well.

PR #662, opened 2023-11-28
Update patches (November 28, 2023)

nathanchance:
Just a couple of changes to drop applied patches, see each commit for the individual details.


JustinStitt (approved):
LGTM!

side note: in the new CI system maybe there's a way to automatically open PR's when patches fail to apply.

PR #663, opened 2023-11-30
patches: Drop SRSO patches from android13-5.10

nathanchance:
These were in 5.10.191 and 5.10.192, which have now been merged into
android13-5.10:

https://android.googlesource.com/kernel/common/+/b9ea98aa2fd47cd8af31bb674d2de84d09251e09


JustinStitt (approved):
thanks for picking this up.

LGTM!

PR #666, opened 2023-12-08
README: Update status badges hourly from external cache

kees:
Instead of having the main website being unreadable due to broken SVG fetches from rate limiting internal to github, grab badge SVG from the hourly updated cache hosted at kernel.outflux.net. These badges additionally don't include the build name in the badge, which makes the matrix much more readable.


nathanchance (approved):
This looks much better to me and if it works around GitHub's completely stupid rate limit, even better :)

JustinStitt (approved):
I can vouch that it looks better. If the external caching works then this is a no-brainer!

nickdesaulniers:
[pull #666 ](https://www.youtube.com/watch?v=WxnN05vOuSM) !!! :smiling_imp: :zombie_man: :coffin:

PR #668, opened 2023-12-19
Temporarily change cron string to test caching

nathanchance:
The caching infrastructure gets out of the way when using
workflow_dispatch, so we need to change the cron string to something
that will run right away when merged. This will be reverted once the job
has started.


nathanchance:
It appears that I merged this too late to matter so I force pushed main with https://github.com/ClangBuiltLinux/continuous-integration2/commit/a37fd93399b6655e27c24380c9e95ed6cf09f7aa for 22:50 UTC. Hopefully that runs properly then I will revert that change.

JustinStitt:
ruh, roh. It's not running.

PR #669, opened 2023-12-20
Fix workflows with caching

nathanchance:
There is a crash when running `update_cache.py` even if `tuxsuite` does not run:

```
  Reading builds.json
    File "update_cache.py", line 88, in <module>
      main()
    File "update_cache.py", line 58, in main
      raw = builds_json.read_text(encoding='utf-8')
    File "/usr/local/lib/python3.8/pathlib.py", line 1236, in read_text
      with self.open(mode='r', encoding=encoding, errors=errors) as f:
    File "/usr/local/lib/python3.8/pathlib.py", line 1222, in open
      return io.open(self, mode, buffering, encoding, errors, newline,
    File "/usr/local/lib/python3.8/pathlib.py", line 1078, in _opener
      return self._accessor.open(self, flags, mode)
  FileNotFoundError: [Errno 2] No such file or directory: 'builds.json'
```

None of the steps after `tuxsuite` need to run, so apply the same condition to all steps.

### DIFF
--- a/.github/workflows/4.14-clang-13.yml
+++ b/.github/workflows/4.14-clang-13.yml
@@ -54,20 +54,25 @@ jobs:
       if: ${{ needs.check_cache.outputs.output == 'success' && github.event_name != 'workflow_dispatch' && needs.check_cache.outputs.status == 'fail'}}
       run: echo 'Cache HIT on previously FAILED build. Failing this build to avoid redundant work.' && exit 1
     - uses: actions/checkout@v4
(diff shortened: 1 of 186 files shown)


JustinStitt:
So if tuxsuite fails to run there is no `builds.json` is this causing the crash? @nathanchance

JustinStitt (approved):
LGTM.

Adding all these checks should avoid running job steps that don't need to run. With the added benefit of dodging a bug with `update_cache.py` when no `builds.json` is present -- since if `tuxsuite` doesn't run we don't get a `builds.json`.

PR #671, opened 2023-12-20
Add Caching Info Section to README

JustinStitt:
Some basics about the system + a diagram.

We can make more or less detailed as we see fit.


nathanchance (approved):
Seems reasonable to me!

Issue #675, opened 2024-01-08
Drop 4.14 (soon)


JustinStitt:
While it isn't [quite time yet](https://endoflife.date/linux), I'm opening this issue so we remember to drop 4.14 from CI in a few weeks (31 Jan 2024).

The introduction of the 6.6 LTS Kernel integration into CI (#673) means we have ramped up our total builds being ran. This will add a new row to the CI matrix, meaning more jobs and more clutter in the matrix itself.

While the CI caching helps here, especially for a kernel as old as 4.14, we should be prompt in dropping our CI support for it nonetheless.

edit:

... and also consider dropping android‑4.14 CI support

nathanchance:
Thanks for opening this! It is actually sooner than you might think, 4.14.336 is out for review and Greg states that it is going to be the last 4.14 release: https://lore.kernel.org/20240108141854.158274814@linuxfoundation.org/

nathanchance:
4.14 is officially EOL with 4.14.336:

https://git.kernel.org/pub/scm/docs/kernel/website.git/commit/?id=8ba05df34af92bc993c6e88393cd71e1516a99c6
https://lore.kernel.org/2024011046-ecology-tiptoeing-ce50@gregkh/
https://git.kernel.org/stable/l/linux-4.14.y

@JustinStitt want to take a crack at dropping 4.14 from CI? You can use the 4.9 EOL pull as a reference, should be mostly the same here: https://github.com/ClangBuiltLinux/continuous-integration2/commit/c3088308043c5c469fcbf3e034228271ed0488a6

JustinStitt:
> @JustinStitt want to take a crack at dropping 4.14 from CI? You can use the 4.9 EOL pull as a reference, should be mostly the same here: https://github.com/ClangBuiltLinux/continuous-integration2/commit/c3088308043c5c469fcbf3e034228271ed0488a6


Yeah, I'll try to tackle that after lunch 👍

PR #679, opened 2024-01-10
drop 4.14

JustinStitt:
drop 4.14 and android-4.14 trees from CI as they are [no longer supported](https://lore.kernel.org/all/2024011046-ecology-tiptoeing-ce50@gregkh/).

We go from 10161 to 10075 builds per week according to `./scripts/estimate_builds.py`

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/675


nathanchance (approved):
LGTM, good job!

According to https://source.android.com/docs/core/architecture/kernel/android-common, it seems like dropping support for `android-4.14-stable` is coming this month, which is presumably due to upstream dropping support for 4.14. It also seems like `android-4.19-stable` will also be unsupported at that time, so maybe we can drop that soon too?

bwendling (approved):

PR #683, opened 2024-01-17
check all builds for metadata

JustinStitt:
Instead of just checking the first build in the builds.json let's check
them all to ensure they have well-formed metadata fields for use in
caching; if this is not the case then we will error out, dumping the
builds.json.


JustinStitt:
Forgot about the ruff lint checks, fixing now.

update_cache.py
```diff
@@ -64,10 +64,25 @@ def main():
...
+                "version_full"]
+        except KeyError:
+            builds_that_are_missing_metadata.append(entry)
+
+    if len(builds_that_are_missing_metadata) > 0:
+        print(
+            "Error: Some of the builds in builds.json are malformed and missing "
+            f"some metadata. Here's a list: {builds_that_are_missing_metadata}\n"
+            f"Here's the build.json in question:\n{raw}")
+        sys.exit(0)
```
bwendling, line 83: Should the exit code be `1` instead of `0`?

update_cache.py
```diff
@@ -64,10 +64,25 @@ def main():
...
+    for entry, build in builds.items():
+        try:
+            git_sha = build["git_sha"]
+            clang_version = build["tuxmake_metadata"]["compiler"][
+                "version_full"]
+        except KeyError:
+            builds_that_are_missing_metadata.append(entry)
+
+    if len(builds_that_are_missing_metadata) > 0:
+        print(
```
bwendling, line 79: The string here is going to have a very long first line. It might be better to print something like this:

```
    print(f"""\
Error: These builds in {raw} are missing metadata:
{builds_that_are_missing_metadata}
""")
```

Think about those poor people who are still using 80-column monitors! :-)

update_cache.py
JustinStitt: yes.

update_cache.py
JustinStitt: `raw` is a giant json dump. `builds_that_are_missing_metadata` is just a list of hashes for the builds with malformed metadata.

Here's some example output (trimmed):


```
[4.14 (clang-13)/TuxSuite (defconfigs)]   🐳  docker exec cmd=[sh -e /var/run/act/workflow/4.sh] user= workdir=
| Reading mock.builds.json
| Error: Some of the builds in builds.json are malformed and missing some metadata. Here's a list: ['2YlyQFeMIAc94Pn5vFMTT2hW7Kt']
| Here's the build.json in question:
| {
|     "builds": {
|         "2YlyQFeMIAc94Pn5vFMTT2hW7Kt": {
|             "auto_retry": true,
|             "build_name": "",
|             "build_status": "pass",
|             "callback": null,
|             "client_token": "23427fe5-0d8c-4df1-b2f5-33d349e29027",
...
(snip)
```

bwendling: Oh...ew...Is that useful for debugging? I guess if this is for a webpage it would be... I'll leave it to your judgement. :-)

bwendling (approved):

update_cache.py
JustinStitt: @nathanchance and I spoke about this and in an event of a failure like what we experienced with: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/7537720415/job/20517094849 we would prefer if the builds.json was dumped. 

So yes, basically only for web usage.

nathanchance (commented):
I am conflicted on whether or not we error out unconditionally on a build being malformed... I think that we only want to error out if we cannot get a `git_sha` or `clang_version` from any builds because that's really what we are after here (and we can still print the malformed builds). Otherwise, we use what we got from one build to update the cache then defer to the check logs stage to error out if a build errored like the ChromeOS ones did. Am I missing something though?

JustinStitt:
@nathanchance The current implementation just checks for `KeyError`s for `git_sha` and `clang_version`. Perhaps I am a bit confused by your suggestion, we should or shouldn't be checking for more fields?

nathanchance:
> @nathanchance The current implementation just checks for `KeyError`s for `git_sha` and `clang_version`

Right but if there are any `KeyError`s, won’t the `update_cache.py` step fail now due to `sys.exit(0)`? That basically leaves us in the exact same situation as we are in now. Apologies that I was not more clear here but I think there should only be a fatal error in `update_cache.py` if we cannot get a SHA or compiler version altogether (as that means something seriously wrong has occurred), not if we find any malformed builds. We don’t need to fail on `update_cache.py`, as the check logs and boot job will fail, which is a more appropriate location to do so in my opinion.

I ultimately envision something like this but I have not tested it:

```python
    builds_that_are_missing_metadata = []
    git_sha = None
    clang_version = None

    for entry, build in builds.items():
        try:
            git_sha = build["git_sha"]
            clang_version = build["tuxmake_metadata"]["compiler"][
                "version_full"]
            break
        except KeyError:
            builds_that_are_missing_metadata.append(entry)
    else:
        raise RuntimeError(“Could not find a suitable git sha or compiler version in all builds”)

    if len(builds_that_are_missing_metadata) > 0:
        print(
            "Error: Some of the builds in builds.json are malformed and missing "
            "some metadata.\n"
            f"Here's a list: {builds_that_are_missing_metadata}\n"
            f"Here's the build.json in question:\n{raw}")
```

update_cache.py
```diff
@@ -64,10 +65,31 @@ def main():
...
+    builds_that_are_missing_metadata = []
+    git_sha = None
+    clang_version = None
+    for entry, build in builds.items():
+        try:
+            git_sha = build["git_sha"]
+            clang_version = build["tuxmake_metadata"]["compiler"][
+                "version_full"]
+        except KeyError:
+            builds_that_are_missing_metadata.append(entry)
```
nathanchance, line 77:
```
    else:
        break
```

so that we don't keep iterating through `builds` once we have a hash and version?

update_cache.py
```diff
@@ -64,10 +65,31 @@ def main():
...
+            f"Could not find a suitable git sha or compiler version in any build\n"
+            f"Here's the build.json:\n{raw}")
+
+    if len(builds_that_are_missing_metadata) > 0:
+        print(
+            "Warning: Some of the builds in builds.json are malformed and missing "
+            "some metadata.\n"
+            f"Here's a list: {builds_that_are_missing_metadata}\n"
+            f"Here's the build.json in question:\n{raw}")
+        sys.exit(0)
```
nathanchance, line 90: Shouldn't this `sys.exit(0)` be removed so that we go and update the cache?

nathanchance (commented):
One minor comment as an optimization and one more potential issue then I think this should be good to go, thanks for sticking with it.

update_cache.py
JustinStitt: Gotcha, checkout [5069c26](https://github.com/ClangBuiltLinux/continuous-integration2/pull/683/commits/5069c260af95a8872a3e181ceb823428c2d5b705)

update_cache.py
JustinStitt: Yes.

[5069c26](https://github.com/ClangBuiltLinux/continuous-integration2/pull/683/commits/5069c260af95a8872a3e181ceb823428c2d5b705)

nathanchance (approved):

PR #685, opened 2024-01-24
Update LLVM tip of tree anchor to 19

nathanchance:
While clang-nightly in TuxSuite has not updated to `clang-18` just yet, it should be soon and this is needed so that updating the generated files works properly.


nathanchance:
Ping for review, this is needed to avoid presubmit failures on other patches (like https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/7634469980).

bwendling (approved):

PR #687, opened 2024-01-25
Decompose generator.yml

nathanchance:
`generator.yml` has become quite long as our matrix has grown, currently coming in at 4458 lines. It is easy to get lost when modifying multiple different trees and LLVM versions.

Decompose `generator.yml` into individual YAML files for each section. We can dynamically rebuild `generator.yml` from all of these pieces in `utils.get_config_from_generator()` and pass that along to `yaml.load()`.

There is currently no functional change. In the future, this will make it easier to add new LLVM versions, as we can copy and modify whole files rather than specific ranges in `generator.yml`.


JustinStitt:
This is great but also creates a lot of fragmentation. It's nice that you numerically ordered these but perhaps we need a checklist (in a README somewhere) about how to add/remove builds and what stuff to update (even in other repositories).

nathanchance:
Agreed. I think I will try to create some maintainer documentation here soon that goes over the structure of the whole repository, what everything does, and how it all fits together with some guides around common manipulations like adding a new tree or LLVM version. Not sure how soon I will be able to get to it but I will definitely prioritize it over the next couple of weeks.

JustinStitt:
OK, with #689 proposed I think this is OK to merge.

JustinStitt (approved):
^^

nathanchance:
The README will be outdated until #689 is handled but I think that's okay. Merging, thanks for the review!

PR #688, opened 2024-01-25
Patch all outstanding failures (Jan 24, 2024)

nathanchance:
See each individual change for the linked issues for more context.


nathanchance:
2e1e313b83d1b60e31a100d3a6568fe4c9c8a202 is new in the latest push, as that failure happened overnight.

JustinStitt (approved):
Yep, LGTM.

I wonder though, if instead of providing the raw patches we could just have lore links that our system could fetch and apply? But as I'm writing this I realize that just complicates things. Food for thought, though.

nathanchance:
I've thought about that as well and came to the same conclusion. It may be nice to eventually have a step that tries to apply the patches first and remove any that fail to apply so that the build does not fail when patches fail to apply like it current does. However, the current scheme makes cleaning up the patches more obvious; if there is no clear signal that a patch no longer applies, we may not clean up the patches in between adding new ones.

bwendling (approved):

JustinStitt:
Merging ASAP since we have +2 review and we could potentially avoid some red in the matrix.

PR #697, opened 2024-01-30
caching: Add hash of contents of patches folder

nathanchance:
Currently, the frontend caching infrastructure is unaware of the patches being applied. This is problematic if we have changed either added, removed, or changed any patches since the previously cached run because the caching would skip the build even when the build's expected outcome would be different than the cached build's result.

Hash the contents of the patches folder using the hashlib library and add that to the cache so that when the patches folder changes at all, the cache is skipped and the build proceeds.

NOTE: I have not formally tested this but it seems like it should be sufficient. Please review carefully.

### DIFF
--- a/caching/check.py
+++ b/caching/check.py
@@ -38,7 +38,7 @@ from typing import Optional
 
 import requests
 
-from utils import get_workflow_name_to_var_name, update_repository_variable
+from utils import get_patches_hash, get_workflow_name_to_var_name, update_repository_variable
 
 OWNER = "ClangBuiltLinux"
 REPO = "continuous-integration2"
@@ -144,13 +144,14 @@ def get_repository_variable_or_none(name: str) -> Optional[dict]:
     return json.loads(as_dict["value"])
 
 
-def create_repository_variable(name: str, linux_sha: str,
-                               clang_version: str) -> None:
+def create_repository_variable(name: str, linux_sha: str, clang_version: str,
+                               patches_hash: str) -> None:
     _url = f"https://api.github.com/repos/{OWNER}/{REPO}/actions/variables"
 
     _value = json.dumps({
         "linux_sha": linux_sha,
         "clang_version": clang_version,
+        "patches_hash": patches_hash,
         "build_status": "presuite",
     })
     data = {"name": name, "value": _value}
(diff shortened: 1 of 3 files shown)


caching/check.py
```diff
@@ -173,11 +174,14 @@ def create_repository_variable(name: str, linux_sha: str,
...
     VAR_NAME = get_workflow_name_to_var_name(args.workflow_name)
+    tree_name = args.workflow_name.split(' ', 1)[0]
 
     curr_sha = get_sha_from_git_ref(args.git_repo, args.git_ref)
     curr_clang_version = get_clang_version()
+    # pylint: disable-next=invalid-name
+    curr_patches_hash = get_patches_hash(tree_name)
     print(
-        f"Current sha: {curr_sha}\nCurrent Clang Version: {curr_clang_version}"
+        f"Current sha: {curr_sha}\nCurrent Clang Version: {curr_clang_version}\nCurrent patches hash: {curr_patches_hash}"
```
bwendling, line 184: For these types of prints, it might be cleaner to do something like this:

```python
    print(f"""\
Current SHA1: {curr_sha}
Current Clang Version: {curr_clang_version}
Current Patches Hash: {curr_patches_hash}
""")

File: caching/check.py [hunk 2]
```diff
@@ -207,18 +212,21 @@ def create_repository_variable(name: str, linux_sha: str,
...
     cached_build_status = result["build_status"]
+    cached_patches_hash = result.get("patches_hash", curr_patches_hash)
 
-    if cached_sha != curr_sha or cached_clang_version != curr_clang_version:
+    if cached_sha != curr_sha or cached_clang_version != curr_clang_version or cached_patches_hash != curr_patches_hash:
         print(
-            f"CACHE MISS: current linux_sha is {curr_sha} and clang_version is {curr_clang_version} "
-            f"while {args.workflow_name} has a cached linux_sha of {cached_sha} "
-            f"and a cached clang_version of {cached_clang_version} under "
+            f"CACHE MISS: current linux_sha is {curr_sha}, clang_version is {curr_clang_version}, "
```
bwendling, line 219: Same comment here about using `f"""`, but also this is a lot of text to wade through. Is it possible to split it up into individual lines like the change above?

caching/check.py
nathanchance: Sure thing, I added 734abf53d0b53d70473901cb7eb65bd3107544bc to accomplish this.

utils.py
```diff
@@ -231,6 +243,8 @@ def update_repository_variable(
             cached_value["linux_sha"] = sha
         if clang_version:
             cached_value["clang_version"] = clang_version
+        if patches_hash:
+            cached_value["patches_hash"] = patches_hash
```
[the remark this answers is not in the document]
nathanchance: Done, thanks!

112aa368029e8bb93e3f1aedc2a8d4ff4aef36b8

bwendling (approved):
Sorry for the delay.

nathanchance:
No worries, thanks for taking a look again!

PR #699, opened 2024-01-30
patches: arm64-fixes: Drop smb __bdos() patch

nathanchance:
It is in 6.8-rc2, which arm64 for-next/fixes is now up to date with.


bwendling (approved):
Approved, but is there any way to place command line options like these in a single place so that you don't have to perform all of these changes?

bwendling:
> > Approved, but is there any way to place command line options like these in a single place so that you don't have to perform all of these changes?
> 
> I am not sure I understand the comment. Do you mean the `--patch` options in the GitHub workflow and TuxSuite files or another place where we process command line options?

I'm just using this change as an example, but it could apply to other places. You remove the `--patch-series` flag here, but if you need to add it back again, you'll have to re-add this.

```shell
tuxsuite plan --git-repo https://git.kernel.org/pub/scm/linux/kernel/git/arm64/linux.git --git-ref for-next/fixes --job-name defconfigs --json-out builds.json --patch-series patches/arm64-fixes tuxsuite/arm64-fixes-clang-11.tux.yml || true
```

It seems like needless churn. In a shell script, one would create variables that held all of the pertinent flags, so that it only needs a change in one place:

```shell
ARM64_PATCH_FLAGS='--patch-series patches/arm64-fixes'

tuxsuite plan ${GIT_REPO} ... ${ARM64_PATCH_FLAGS} ...
```

Is something like this at all possible with YAML?

nathanchance:
It may be possible but these files are all automatically generated anyways, so doing an optimization like that does not necessarily seem super useful to me. I don't usually look at the diff of the generated files much beyond just making sure the changes seem expected.

PR #701, opened 2024-01-30
Attempt to rebalance builds

nathanchance:
As the number of supported trees and toolchain combinations grow, there is a possibility of heavy contention when a large number of builds are started at once.

To help avoid this, start to rebalance the builds by breaking apart certain anchors like the Monday/Friday, Wednesday, and Sunday anchors to have a variety of start times.

I would like to add other days eventually but I figured I would start smaller at first. This lays the groundwork for doing that later.


JustinStitt (approved):
Yep. This is a good idea. A lot of the builds all try to run at the same time just bogging down GitHub Actions with some builds being placed in a pending state.

LGTM.

PR #702, opened 2024-01-31
Temporarily disable LLVM tip of tree builds for RISC-V

nathanchance:
TuxMake temporarily disabled RISC-V support with clang-nightly because apt.llvm.org currently does not have LLVM 19 builds for the unstable distribution, which the RISC-V container uses as a base.

While this issue gets sorted, drop RISC-V builds with clang-nightly, as that combination will error now.

Link: https://lore.kernel.org/20240131174445.GA535345@dev-arch.thelio-3990X/


JustinStitt:
Should we just let it error out? What is the timeline on the support being added? I'm thinking of a scenario where we merge this and it instantly is fixed and we need a revert within the same day. FWIW, the changes themselves match what I'd expect from dropping some jobs 👍

nathanchance:
This is likely going to be dealt with relatively quickly, so it's probably okay to hold off on merging this. The only thing I worry about is TuxSuite erroring before any other builds run, meaning this kills our `clang-nightly` testing altogether. I have no strong opinion though.

nathanchance:
This is now resolved and no longer needed.

PR #703, opened 2024-01-31
Update {upload,download}-artifact actions to v4

nathanchance:
This is needed to avoid warnings around Node 16 actions being deprecated.


JustinStitt (approved):
LGTM.

Let's keep an eye out on regressions, though. There's a list of breaking changes from the v3->v4 upgrade available here:
https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/

nathanchance:
I should have noted that I do not think any of those caveats apply as we already use a unique `name` per workflow... Thanks for taking a look!

Issue #710, opened 2024-02-14
Add support for building s390 with `LLVM=1`


nathanchance:
LLD has gained s390 support in main (and hopefully 18.1.0):

https://github.com/llvm/llvm-project/commit/fe3406e349884e4ef61480dd0607f1e237102c74
https://github.com/llvm/llvm-project/pull/81675

There were several issues uncovered on the Linux side that needed to be fixed:

https://github.com/ClangBuiltLinux/linux/issues/1991
https://github.com/ClangBuiltLinux/linux/issues/1992
https://github.com/ClangBuiltLinux/linux/issues/1993
https://github.com/ClangBuiltLinux/linux/issues/1994
https://github.com/ClangBuiltLinux/linux/issues/1995

Once all of those changes are available in -next and apt.llvm.org has been updated to include that LLVM change, we should cut all of the relevant LLVM builds on -next over to linking with ld.lld. Ideally, we would just cut over to `LLVM=1 OBJCOPY=s390x-linux-gnu-objcopy OBJDUMP=s390x-linux-gnu-objdump`, as that is known to work with all of the configurations that we test.

nathanchance:
I have updated the title to reflect that:

- `llvm-objcopy` works with the latest LLVM `release/18.x` (https://github.com/llvm/llvm-project/commit/5ef297ab611822537a385d45244867519563d3ef) and `main` (https://github.com/llvm/llvm-project/commit/3c02cb7492fc78fb678264cebf57ff88e478e14f) branches and the current s390 features branch (https://git.kernel.org/s390/c/9ea30fd166e9b869546059f714beab78cb150b11)
- `llvm-objdump` will work with the s390 features branch once https://lore.kernel.org/linux-s390/20240220-s390-work-around-llvm-objdump-t-j-v1-1-47bb0366a831@kernel.org/ has been pushed out (https://lore.kernel.org/linux-s390/20240221091612.7561-A-hca@linux.ibm.com/)

With this, we can cut over to full `LLVM=1` with `ARCH=s390` with Linux trees and LLVM toolchains that satisfy these conditions (we will have to validate the versions of LLVM that `tuxsuite` has available).

nathanchance:
https://github.com/ClangBuiltLinux/continuous-integration2/pull/712

PR #711, opened 2024-02-16
Add support for linking certain s390 kernels with ld.lld

nathanchance:
s390 support in `ld.lld` has recently landed in the main branch of LLVM. With this, it makes sense to start cutting s390 over to full `LLVM=1`, versus the current situation of just `CC=clang`. linux-next is the only tree that currently has the necessary changes to support `ld.lld` but they will trickle down to other trees in time.

`llvm-objcopy` currently works on LLVM main as well but that needs a kernel side fix to boot properly, so we will stick with GNU objcopy until that is fixed.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/710
Link: https://github.com/ClangBuiltLinux/linux/issues/1996
Link: https://github.com/llvm/llvm-project/commit/fe3406e349884e4ef61480dd0607f1e237102c74
Link: https://github.com/llvm/llvm-project/commit/3c02cb7492fc78fb678264cebf57ff88e478e14f

### DIFF
--- a/.github/workflows/next-clang-19.yml
+++ b/.github/workflows/next-clang-19.yml
@@ -890,12 +890,12 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: scripts/check-logs.py
-  _a7016a4f3a73a508d1b00dd9b9dfa168:
+  _f97d7f8efcb2c842d35f0e22f01a52bf:
     runs-on: ubuntu-latest
     needs:
     - kick_tuxsuite_defconfigs
     - check_cache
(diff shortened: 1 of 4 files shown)


JustinStitt:
talked offline: waiting on llvm-objcopy to hit this PR then we're good to go!

nathanchance:
That change has been applied (but not pushed out just yet): https://lore.kernel.org/linux-s390/20240221091612.7561-A-hca@linux.ibm.com/

I am going to close this pull request and redo it with a full cut over to `LLVM=1` and I have reflected that in https://github.com/ClangBuiltLinux/continuous-integration2/issues/710.

PR #713, opened 2024-02-21
badge-caching: Use README.md as source of truth

kees:
Add an additional "unknown" state for if/when parsing of a fetched SVG fails to find either "passing" nor "failing".

Parse the README.md for yml URLs.

Log SVG scraping and ignore failures.

### DIFF
--- a/scripts/badge-caching/refresh.sh
+++ b/scripts/badge-caching/refresh.sh
@@ -4,10 +4,19 @@ set -e
 BASE="/var/www/html/cbl/badges"
 cd "$BASE"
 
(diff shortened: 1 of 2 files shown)


nathanchance (approved):
LGTM, we should still think about how to avoid the duplication between this README and clangbuiltlinux.github.io but I think this is a good direction to move in regardless.

JustinStitt (approved):
LGTM, ran the script locally.

nickdesaulniers:
> LGTM, we should still think about how to avoid the duplication between this README and clangbuiltlinux.github.io

Perhaps a git submodule that's included in two places? Dunno if that works for GitHub pages though...

PR #719, opened 2024-02-29
Use kernel.org clang whenever possible

nathanchance:
This moves our builds away from Debian's LLVM to the kernel.org LLVM builds I maintain, which are faster and within our control for updates.

I would like to do this for `clang-nightly` eventually but I have to make sure it is maintainable before doing so, so this just does the stable LLVM builds for now.

### DIFF
--- a/.github/workflows/4.19-clang-13.yml
+++ b/.github/workflows/4.19-clang-13.yml
@@ -19,7 +19,7 @@ jobs:
   check_cache:
     name: Check Cache
     runs-on: ubuntu-latest
-    container: tuxmake/x86_64_clang-13
+    container: tuxmake/x86_64_korg-clang-13
     env:
       GIT_REPO: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
       GIT_REF: linux-4.19.y
--- a/.github/workflows/4.19-clang-14.yml
+++ b/.github/workflows/4.19-clang-14.yml
@@ -19,7 +19,7 @@ jobs:
   check_cache:
     name: Check Cache
     runs-on: ubuntu-latest
-    container: tuxmake/x86_64_clang-14
+    container: tuxmake/x86_64_korg-clang-14
     env:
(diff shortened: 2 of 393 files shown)


sylvestre:
oh :(

Why faster? pgo/lto?

nathanchance:
> oh :(

You have done a great job with apt.llvm.org and I continue to recommend it for people who want continuous updates automatically from their package manager. This pull request is not a reflection of that at all.

> Why faster? pgo/lto?

Yes, I apply PGO, ThinLTO, and BOLT whenever possible, which can result in 50%+ speed up at run time.

https://gitlab.com/Linaro/tuxmake/-/issues/215

https://mirrors.edge.kernel.org/pub/tools/llvm/

sylvestre:
> You have done a great job with apt.llvm.org and I continue to recommend it for people who want continuous updates automatically from their package manager. This pull request is not a reflection of that at all.

thanks

> Yes, I apply PGO, ThinLTO, and BOLT whenever possible, which can result in 50%+ speed up at run time.

I could have enabled these for you if you had asked ;)

nickdesaulniers:
> > Yes, I apply PGO, ThinLTO, and BOLT whenever possible, which can result in 50%+ speed up at run time.
> 
> I could have enabled these for you if you had asked ;)

Where is the best venue to formally ask? llvm's issue tracker, another bug tracker, email?

nathanchance:
> > Yes, I apply PGO, ThinLTO, and BOLT whenever possible, which can result in 50%+ speed up at run time.
> 
> I could have enabled these for you if you had asked ;)

I think it is still worth doing even in spite of this pull request (I maintain those toolchains for other developers, not just our CI), especially for the main Debian packages because I know many kernel developers who use Debian and Ubuntu clang versions and I am sure they would enjoy those speed ups :)

nickdesaulniers (approved):

sylvestre:
> Where is the best venue to formally ask? llvm's issue tracker, another bug tracker, email?

https://github.com/llvm/llvm-project/issues/new for example :)

PR #720, opened 2024-03-01
generator: workflow: Use setup-python action in check cache stage

nathanchance:
Recently, the check cache stage started failing with:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.
```

This occurs because the container this command is being run in got upgraded silently from Debian 11 to Debian 12, where the newer Python version allows distributors to block using pip on the system level by default.

To avoid this altogether, use the setup-python action to override the system Python. Alternative solutions are using a virtual environment like the command suggests or using '--break-system-packages'. However, this ensures that we always have access to a consistent environment and a modern Python version, in case we ever want to take advantage of newer language features.

### DIFF
--- a/.github/workflows/4.19-clang-13.yml
+++ b/.github/workflows/4.19-clang-13.yml
@@ -28,12 +28,15 @@ jobs:
       status: ${{ steps.step2.outputs.status }}
     steps:
     - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
     - name: pip install -r requirements.txt
-      run: apt-get install -y python3-pip && pip install -r requirements.txt
-    - name: python check_cache.py
+      run: python3 -m pip install -r requirements.txt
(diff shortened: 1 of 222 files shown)


nathanchance:
I am going to merge this without any review because this is hard blocking all jobs but `clang-nightly` ones. If there are any major issues, I am happy to resolve them with a follow up PR.

PR #730, opened 2024-03-15
Update patches (March 15, 2024)

nathanchance:
See the individual commits for more details.


JustinStitt (approved):
Ya looks good. I wish there was a better way to do this, though.

Issue #732, opened 2024-03-19
caching/update.py failing because Tuxmake metadata is empty


JustinStitt:
We're getting `RuntimeError: Could not find a suitable git sha or compiler version in any build` during the Update cache phase.

This is probably because the `tuxmake_metadata` field is empty. We can see that here: https://github.com/ClangBuiltLinux/continuous-integration2/actions/runs/8335729494/job/22811740883

Why is this field empty for some builds?

edit: It seems this is caused when a patch fails to apply. I never noticed this exact failure in that instance though. I will look to refine this error path with better handling/messaging and then close this issue.

JustinStitt:
Opened a PR here: https://github.com/ClangBuiltLinux/continuous-integration2/pull/733. closing this issue.

PR #733, opened 2024-03-19
add better error handling for patches that fail to apply

JustinStitt:
I noticed that when patches fail to apply we don't get the best error message from `caching/update.py`:

```
...
RuntimeError: Could not find a suitable git sha or compiler version in any build
...
```

I wrote about this in #732.

Anyways, now we'll get an error message like this:
```
| Error: Some patches failed to apply.
| ['35f20786c481d5ced9283ff42de5c69b65e5ed13.patch']
```

Then, we can even have issues or PRs created auto-magically https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#create-an-issue.

Following that, it would just take a human reviewer to say "Yep, this patch has already been applied which has caused this failure; time to merge this robot's PR! I love robots 🤖"

### DIFF
--- a/caching/update.py
+++ b/caching/update.py
@@ -93,17 +93,19 @@ def main():
             build_log_raw = response.read().decode()
 
         failed_pattern = (
-            r"(?<=Apply patch set FAILED\s)[0-9A-Za-z._:/\-\s]*?(?=\serror: )"
-        )
+            r"(?<=Apply patch set FAILED\s)[0-9A-Za-z._:/\-\s]*?(?=\serror: )")
         failed_matches = re.findall(failed_pattern, build_log_raw)
         if len(failed_matches) == 0:
             print(
                 f"No patches failed to apply yet the build status stated there were: {build['status_message']}"
             )
-            sys.exit(0)  # Not sure how we got here but continue the action anyways
+            sys.exit(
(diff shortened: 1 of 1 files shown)


caching/update.py
```diff
@@ -79,6 +81,33 @@ def main():
...
+
+        failed_pattern = (
+            r"(?<=Apply patch set FAILED\s)[0-9A-Za-z._:/\-\s]*?(?=\serror: )")
+        failed_matches = re.findall(failed_pattern, build_log_raw)
+        if len(failed_matches) == 0:
+            print(
+                f"No patches failed to apply yet the build status stated there were: {build['status_message']}"
+            )
+            sys.exit(
+                0)  # Not sure how we got here but continue the action anyways
```
JustinStitt, line 103: Yeah, I hate this formatting too. But this is what had to be done to appease the linter gods I guess. I'm pretty sure this is from `$ black caching/update.py`

nathanchance: I don't really care when `yapf` or other formatters mess with formatting and make it somewhat wonky. I think moving the comment up a line to fix this is probably worth it since it is going to split it across two lines anyways. Up to you though!

nathanchance (approved):
LGTM, thanks!

PR #734, opened 2024-03-19
Disable MIPS builds that use GNU ld

nathanchance:
After the upgrade from binutils 2.35.2 in Debian bullseye to binutils 2.40 in Debian bookworm, there is a build error:

```
llvm-objcopy: error: Link field value 24 in section .rel.dyn is not a symbol table
```

This is resolved in binutils 2.41 with commit [6b958fe36b7](https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=6b958fe36b765f70878e8d3d002864967c4bc3a4) ("Setting sh_link for SHT_REL/SHT_RELA"). Once Debian has binutils 2.41 or newer, these builds can be re-enabled.

### DIFF
--- a/.github/workflows/5.10-clang-11.yml
+++ b/.github/workflows/5.10-clang-11.yml
@@ -246,34 +246,6 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: scripts/check-logs.py
-  _772f9a8c79e52c6ee000b01cab13b4f6:
-    runs-on: ubuntu-latest
-    needs:
-    - kick_tuxsuite_defconfigs
-    - check_cache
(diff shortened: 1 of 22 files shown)


JustinStitt (commented):
Depending on how long it takes Debian to update, I'm fine with just letting the MIPS builds fail. Inevitably, we'll have to revert this PR.

nathanchance:
It may be possible to ask the TuxMake folks to update the base of the images to Debian Sid for potential access to a newer version (although I have not checked) but that can cause more tangential issues because Sid might be broken, so I don't think it is wise.

Debian won't be updated until summer of 2025, so I don't think we can realistically let this be broken for that long :)

JustinStitt (approved):

PR #746, opened 2024-04-26
patches: Drop Bluetooth patch from -tip

nathanchance:
This change is now in tip/master via Linus's tree.


bwendling (approved):

JustinStitt:
Hi, sorry I'm late here. (manual ping @nathanchance just in case you have notifications off for merged stuff)

I wonder if the run lines could keep the `--patch-series` line so the introduction or removal of patches only touches the `series` file in the respective folder.

Imagine if dropping a patch for 6.6 looked like this:
```diff
diff --git a/patches/6.6/series b/patches/6.6/series
index bfa1b61..fdcfafa 100644
--- a/patches/6.6/series
+++ b/patches/6.6/series
@@ -1,2 +1 @@
-9bf4e919ccad613b3596eebf1ff37b05b6405307.patch
 v2_20240305_nathan_media_mxl5xx_move_xpt_structures_off_stack.patch
```
... instead of 15+ yml files being touched to add or remove `--patch-series`!

nathanchance:
Yeah, I suppose we could just force add empty `series` files for every tree and just add `--patch-series` to `tuxsuite` always. At the same time, that would make it much more difficult to see what trees have a patch applied to them by just looking at the patches folder.

PR #750, opened 2024-05-02
Add a CONFIG_CFI_CLANG build for ARCH=arm on -next with LLVM >= 16

nathanchance:
`CONFIG_CFI_CLANG` is [now supported for `ARCH=arm` on -next](https://git.kernel.org/next/linux-next/c/1a4fec49efe5273eb2fcf575175a117745f76f97). kCFI was first added in LLVM 16, so add an `ARCH=arm defconfig` + `CONFIG_CFI_CLANG=y` build when using LLVM 16 or newer.


kees (approved):
This looks good to me! Yay coverage!

PR #753, opened 2024-05-15
scripts/build-local.py: Add else with assertion to clear up new pylint warning

nathanchance:
```
scripts/build-local.py:167:43: E0606: Possibly using variable 'distro' before assignment (possibly-used-before-assignment)
```

### DIFF
--- a/scripts/build-local.py
+++ b/scripts/build-local.py
@@ -164,6 +164,8 @@ for name, builds in jobs.items():
                 distro = 'fedora'
             elif 'openSUSE' in url:
                 distro = 'opensuse'
+            else:
+                raise RuntimeError(f"URL ('{url}') was not handled?")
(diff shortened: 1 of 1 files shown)


scripts/build-local.py
```diff
@@ -164,6 +164,8 @@ def interrupt_handler(_signum, _frame):
                 distro = 'fedora'
             elif 'openSUSE' in url:
                 distro = 'opensuse'
+            else:
+                raise RuntimeError("URL ('{url}') was not handled?")
```
JustinStitt, line 168: This should be an f-string right?

`f"URL ('{url}') was not handled?"`

nathanchance: Yes, thank you for catching this! Fixed on the force push.

JustinStitt (approved):

PR #764, opened 2024-07-19
patches: mainline: Drop radeon __counted_by patch

nathanchance:
It has been merged: https://git.kernel.org/linus/2bac084468847cfe5bbc7166082b2a208514bb1c


nathanchance:
I have updated this to include the removal of the same patch from -tip.

PR #771, opened 2024-08-12
Add LLVM 19 builds

nathanchance:
This moves the current LLVM latest builds to LLVM 18 and moves LLVM latest builds to LLVM 19. This results in a net addition of 1200 builds per week.

```
-Total builds per week: 10060
+Total builds per week: 11279
```

See the individual commits for more details.


JustinStitt:
Nice, the increase in builds is not too big of a deal. Although, I do want to add some stats to how many builds we are cache-hitting; I'll look into this.

JustinStitt (approved):

PR #787, opened 2024-11-18
patches: Apply fix for -Wframe-larger-than in vdec_vp9_req_lat_if.c  

nathanchance:
As there has been no movement on [the upstream report](https://github.com/llvm/llvm-project/issues/111903), apply Arnd's workaround, which I have [requested to be applied](https://lore.kernel.org/20241118200641.GA768549@thelio-3990X).


JustinStitt (approved):
for you: thanks for staying on top of fixes like these 👍 

for the PR: yeah this looks like the others that have went through. LGTM

Issue #796, opened 2024-12-16
Explore adding a check that patches apply at current revision before calling `tuxsuite`


nathanchance:
Now that we have a step that runs before `tuxsuite` (`check_cache`), we should explore adding another step that checks that any patches we need to apply for that build will apply cleanly. This would make patch application failures appear more obvious and avoid blowing up `tuxsuite`.

PR #800, opened 2025-01-02
Add a step to check if patches apply before invoking TuxSuite

nathanchance:
There is no point in spinning up TuxSuite for a build if any patches that we need to apply are going to fail to do so. At the beginning of a build, check if the patches apply to the current repository and reference. This is done by downloading a tarball of the source, running `git init` and committing the files, then finally running `git quiltimport` to make sure that that patches still apply. It would be simpler to use `git clone --depth=1` but cloning can be expensive on the server side and these jobs may run at the same time.

Closes: https://github.com/ClangBuiltLinux/continuous-integration2/issues/796

NOTE: I have not bumped the generated files to make this easier to review and reason through. I will do it before merging.

### DIFF
--- a/scripts/check-patch-apply.py
+++ b/scripts/check-patch-apply.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+# pylint: disable=invalid-name
+
+from argparse import ArgumentParser
+from pathlib import Path
+import os
+import subprocess
+import sys
+from tempfile import TemporaryDirectory
+
+parser = ArgumentParser(
+    description='Check that patches apply to their tree before running CI')
+parser.add_argument(
+    '-p',
+    '--patches-dir',
+    help='Path to patches directory (can be relative or absolute)',
+    required=True,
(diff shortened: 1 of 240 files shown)


scripts/check-patch-apply.py
```diff
@@ -0,0 +1,109 @@
...
+    print(f"{patches_dir} does not contain any patches?")
+    sys.exit(1)
+
+# Rather that invoke 'git clone', which can be expensive for servers depending
+# on the frequency and duration of requests, we fetch a tarball and 'git init'
+# that.
+with TemporaryDirectory() as workdir:
+    # Fetch the tarball from the repository. This is different for each type of
+    # tree that we support.
+    if args.repo.startswith(
```
JustinStitt, line 46: Looks like a good spot for Python's new-ish `match` expression. I'm always looking for an excuse to use it.

Not a blocker for approval though.

JustinStitt (approved):

PR #803, opened 2025-01-20
generator: Run 'apt-get update' before 'apt-get install'

nathanchance:
Over the past few days, the `pip install` step of the `check_cache` job has been failing intermittently with:

```
After this operation, 3170 kB of additional disk space will be used.
Get:1 http://deb.debian.org/debian bookworm/main amd64 python3-pip-whl all 23.0.1+dfsg-1 [1717 kB]
Err:2 http://deb.debian.org/debian bookworm/main amd64 python3-setuptools-whl all 66.1.1-1
  404  Not Found [IP: 151.101.202.132 80]

While it may not be strictly necessary, this can sometimes be avoided by updating the package list before installing any packages. Be a little more defensive to try and avoid failing during the check cache stage.

### DIFF
--- a/.github/workflows/4.19-clang-13.yml
+++ b/.github/workflows/4.19-clang-13.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: pip install -r requirements.txt
-      run: apt-get install -y python3-venv && python3 -m venv venv && . venv/bin/activate && pip install -r requirements.txt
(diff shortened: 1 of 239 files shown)


JustinStitt (approved):
Do you use [act](https://github.com/nektos/act) to test workflows locally?

LGTM.

nathanchance:
> Do you use [act](https://github.com/nektos/act) to test workflows locally?

No but I should definitely look into it (albeit maybe in a virtual machine so I do not have to install `docker` on my host).

PR #806, opened 2025-01-22
scripts/check-patches-apply.py: Fix strip value for refs with a slash

nathanchance:
For refs with a slash in them, such as the arm64 refs `for-next/core` and `for-next/fixes`, the tarball will be formatted like:

```
for-next/core/Makefile
for-next/core/arch/...
```

which causes an incorrect layout with the current value of 1 and patches that would normally apply correctly fail to do so. The strip value for the above scenario needs to be 2 to allow the patches to apply correctly. Use the length of the list created by splitting on slashes as the strip value so that the layout is consistent for patch application.


nathanchance:
Merging because our arm64 builds have not run for almost a week.

PR #807, opened 2025-01-22
Disable ARM configs with BPF enabled in -next with clang-13

nathanchance:
After commit [9aa0ebde0014](https://git.kernel.org/bpf/bpf-next/c/9aa0ebde0014f01a8ca82adcbf43b92345da0d50) ("bpf, verifier: Improve precision of BPF_MUL"), there is an error in certain ARM configurations that enable `CONFIG_BPF_SYSCALL`:

```
ld.lld: error: undefined symbol: __mulodi4
>>> referenced by verifier.c:14221 (/builds/linux/kernel/bpf/verifier.c:14221)
>>>               kernel/bpf/verifier.o:(adjust_reg_min_max_vals) in archive vmlinux.a
>>> referenced by verifier.c:14222 (/builds/linux/kernel/bpf/verifier.c:14222)
>>>               kernel/bpf/verifier.o:(adjust_reg_min_max_vals) in archive vmlinux.a
>>> referenced by verifier.c:14223 (/builds/linux/kernel/bpf/verifier.c:14223)
>>>               kernel/bpf/verifier.o:(adjust_reg_min_max_vals) in archive vmlinux.a
>>> referenced 1 more times
```

This was [encountered previously](https://github.com/ClangBuiltLinux/linux/issues/1438), where it was fixed in clang-14 and avoided in the kernel with a source code workaround (that ended up being cleaner anyways).

This time around, inserting a source code workaround would not be as clean, as it may involve disabling a core part of the kernel on a limited condition (as it only impacts one supported LLVM version and architecture combination) or having a separate code path for this situation.

For now, just disable the builds that are impacted by this. If more people notice this problem, we can explore bumping the minimum supported version of LLVM for building `ARCH=arm` to 14.

### DIFF
--- a/.github/workflows/mainline-clang-13.yml
+++ b/.github/workflows/mainline-clang-13.yml
@@ -117,35 +117,6 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: scripts/check-logs.py
-  _3dcc2dad9e1ef4e6a3f358ffef73f7fe:
-    runs-on: ubuntu-latest
-    needs:
-    - kick_tuxsuite_defconfigs
-    - check_cache
-    - check_patches
-    name: ARCH=arm LLVM=1 LLVM_IAS=1 LLVM_VERSION=13 aspeed_g5_defconfig
-    if: ${{ needs.check_cache.outputs.status != 'pass' }}
-    env:
-      ARCH: arm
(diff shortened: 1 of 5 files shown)


nathanchance:
Mainline needs this now as well.

nickdesaulniers (approved):

PR #809, opened 2025-01-25
Apply .ARM.attributes patch to fix orphan section errors

nathanchance:
See the commit message of the patch file for a full description of the error that necessitates applying it.


nathanchance:
I pushed an update for the chromeos and android branches that will need this.

PR #818, opened 2025-02-21
Drop x86 builds with LLVM 13 and 14 on -next and tip

nathanchance:
[LLVM 15 is becoming the minimum supported version of LLVM for building x86 kernels](https://lore.kernel.org/20250220-x86-bump-min-llvm-for-stackp-v1-1-ecb3c906e790@kernel.org/), so drop these builds from the matrix, as they already do not build.


nathanchance:
Merging, as my patch is in -tip and -next: https://git.kernel.org/tip/7861640aac52bbbb3dc2cd40fb93dfb3b3d0f43c

PR #823, opened 2025-04-03
Add wcslen() patch series to every tree that needs it

nathanchance:
See the individual patch files for the reason this is needed.


nathanchance:
Linus rejected this solution at pull time upstream so I am closing this pull request. I will open a new pull request with the version of the patch he would prefer.

PR #825, opened 2025-04-07
Add android16-6.12

nathanchance:
This is the latest Android Common Kernel branch for the most recent LTS release. As this branch only supports clang-19 (as the release of AOSP LLVM used on that branch) and newer, only wire up support for those same versions in our matrix.


nickdesaulniers:
@adelva1984 can you confirm?

kees:
Is this still desired?

PR #826, opened 2025-04-08
Add wcslen() patch to every tree that needs it (April 7, 2025)

nathanchance:
See the individual patch files for the reason this is needed.

This is the same as #823 but with the solution that Linus would prefer.


nathanchance:
Linus merged this himself in commit [84ffc79bfbf7](https://git.kernel.org/linus/84ffc79bfbf70c779e60218563f2f3ad45288671) ("kbuild: Add '-fno-builtin-wcslen'"), so I have dropped applying this change to mainline and tip.

nickdesaulniers (approved):

PR #831, opened 2025-05-03
Initial matrix reduction

nathanchance:
See the individual commits for more details but at a high level, this pull request aims to try and reduce the number of builds we are doing due to potential resource constraints, while not taking a massive hit in the actual results of CI. Namely, it eliminates several trees that are low value or have alternatives for coverage, reduces builds done with older toolchains, and reduces the mainline build frequency. All in all, this kills about 20% of the matrix.


kees (approved):
That's a lot of builds saved per week! Nice work. This looks good to me.

PR #896, opened 2025-12-23
Disable i386 builds with clang < 17 on 6.18 and newer

nathanchance:
Commit [e2ffa15b9baa](https://git.kernel.org/linus/e2ffa15b9baa447e444d654ffd47123ba6443ae4) ("kbuild: Disable CC_HAS_ASM_GOTO_OUTPUT on clang < 17") broken i386 by exposing a path that was not built previously (see the linked CBL issue). There does not appear to be an easy fix for this, so for now, just disable these builds to keep the matrix green and see if anyone else actually cares about this broken combination. It should be easy enough for folks to upgrade their compiler if they rely on `ARCH=i386` building correctly.

### DIFF
--- a/.github/workflows/mainline-clang-15.yml
+++ b/.github/workflows/mainline-clang-15.yml
@@ -494,35 +494,6 @@ jobs:
         name: boot_utils_json_defconfigs
     - name: Check Build and Boot Logs
       run: scripts/check-logs.py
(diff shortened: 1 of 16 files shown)


kees (approved):
Yup, seems a reasonable way to go to me.

PR #929, opened 2026-06-06
Drop clang-15 and clang-16 builds from mainline, next, and tip

nathanchance:
The series to bump the minimum supported version of LLVM for building the kernel to 17.0.1 has landed in linux-next, so drop the builds there, as they error during the configuration stage now.

While this series is not in mainline or tip yet, it will be soon and the builds have been broken since the 7.1 merge window, which was one of the catalysts for the bump, so drop these as well.


nickdesaulniers (approved):
LGTM; thanks. For posterity, got a link to _what_ necessitated the version bump?

nathanchance:
> For posterity, got a link to _what_ necessitated the version bump?

Sure thing, I tried to give a solid primer in the commit message of the bump:

https://git.kernel.org/kbuild/c/ce3267a39a92be732429fa1a1f6d95a807a31fa7

nickdesaulniers:
Thanks! Wonderfully details kernel commit message.  Sad to see my fingerprints on those fixes, lol! That was a lifetime ago

PR #930, opened 2026-06-16
Update stable anchor to 7.1

nathanchance:
This is the latest stable release.

Apply commit 141848cf ("generator: yml: Drop clang-15 and clang-16 builds from mainline, next, and tip") to stable.


nickdesaulniers:
The commit description on the base commit (5cd6046aa6e9066b94613cf306c6d30bd8cce626) mentions 7.0 when it should mention 7.1.

nathanchance:
Whoops, good catch! Should be fixed now.

nickdesaulniers (approved):

PR #932, opened 2026-06-22
Update actions versions to their latest major version

nathanchance:


nickdesaulniers:
sorry I missed this; have been on vacation the past 2 weeks. Just getting back online today

PR #933, opened 2026-07-08
Further build resource conservation (July 8, 2026)

nathanchance:
Each change should be fairly self explanatory.


nickdesaulniers (approved):
Hopefully not too much delta between Android GKI kernels and LTS that would be relevant here.  If we see breakage downstream we can discuss bringing these builds back.

PR #939, opened 2026-08-14
Update setup-uv to v10.0.1

nathanchance:


nickdesaulniers (changes requested):
Just curious, this is a lot of churn for a very minor version bump. Is it worth it? Do you have something that's going around and auto-updating deps?

nathanchance:
> Is it worth it?

What does it cost? The generated files are basically write-only, so I do not really see this as expensive, aside from just asking for reviews.

> Do you have something that's going around and auto-updating deps?

No, I am just subscribed to the workflow repositories for releases then I go manually bump them.

Technically, the only reason these updates happen frequently for `setup-uv` is because back in `v8.0.0`, they stopped allowing `@v8`, `@v9`, etc to pick the latest stable release within that series, instead requiring the full tag. I can just start ignoring minor releases and only bumping the major version when it changes if you would prefer that.

JustinStitt:
@nickdesaulniers 
> Just curious, this is a lot of churn for a very minor version bump. Is it worth it? Do you have something that's going around and auto-updating deps?

I mean, updating things is probably good most of the time.

The reason this diff is so large is because every workflow file needs its own little update. Is there some "stub" system that github actions has?

```
$ cat shared-stub.yml
...whatever...
      with:
        name: boot_utils_json_defconfigs
      - uses: astral-sh/setup-uv@v10.0.1

$ cat 5.10-clang-17.yml
%!&shared-stub.yml&!%
...whatever...

$ cat 5.10-clang-18.yml
%!&shared-stub.yml&!%
...whatever...
```

you get the idea.

nickdesaulniers:
> What does it cost?

My time.  It looks like I've been the one reviewing the past few PRs here (at least the last 4 I looked at just now).  Perhaps @JustinStitt can help pick up more reviews in this repo?

> I can just start ignoring minor releases and only bumping the major version when it changes if you would prefer that.

Unless something is actively broken or causing the CI to be red, I wouldn't bother keeping these dependencies up to date for every patch level revision.  What if they have patch level releases every week? Waste of time.

> I mean, updating things is probably good most of the time.

I agree.  When is it _not_ good?  When we're bumping for every patch-level or minor-level release, IMO.

> The reason this diff is so large

That's not my problem with this PR.

JustinStitt:
`continuous-integration3 - Revenge of the CI Bill`  is coming soon. maybe we can conjure up 
a new system that requires less overall churn.

nathanchance:
> > What does it cost?
> 
> My time. It looks like I've been the one reviewing the past few PRs here (at least the last 4 I looked at just now).

Sure, that is fair. I don't want to take reviewers' time for granted but I do think there is a difference between a substantial vs. mechanical change in terms of how necessary it is to review a change in the first place. Let's be honest, I could have just pushed this to `main` and not bothered anyone.

> > I can just start ignoring minor releases and only bumping the major version when it changes if you would prefer that.
> 
> Unless something is actively broken or causing the CI to be red, I wouldn't bother keeping these dependencies up to date for every patch level revision. What if they have patch level releases every week? Waste of time.

Alright, I can just monitor the builds and release notes to see if there is anything relevant for us in those releases before bumping them here.

That said, I don't really buy the "patch level releases every week? Waste of time" part of your argument. Maybe it is in this case just because it is a basic GitHub Actions workflow but I don't think it has to be, as I noted above with needing to review mechanical vs. substantial changes. You could argue upgrading to the latest stable Linux version is a waste of time if you haven't had any problems ;)

> `continuous-integration3 - Revenge of the CI Bill` is coming soon. maybe we can conjure up a new system that requires less overall churn.

I am working on a proof of concept right now but if we want to keep the `generator` part of this (and I suspect we do), I am not sure that will change.

